PR #141, opened 2022-04-13
#3694879: Add inverse link version

veej:
Closes [#3694879](https://buildo.kaiten.io/3694879)

### tests performed


packages/storybook/stories/theme.css.ts
```diff
@@ -6,7 +6,9 @@ const pixelToRem = (px: number) => `${px / remBaseSize}rem`;
 
 const colors = {
   white: "#FFFFFF",
+  white_80: "rgba(255, 255, 255, 0.8)",
```
veej, line 9: unrelated, ma ne ho approfittato per fare un po' di pulizia tra tutti i colori con alpha

packages/bento-design-system/src/util/TextChildren.tsx
```diff
@@ -105,8 +112,8 @@ export function textChildrenToChildrenArray(
       return [
         <Box
           as={LinkComponent}
-          className={clsx(linkStyle, bentoSprinkles({ fontWeight: "label" }))}
-          textDecoration={linkConfig.labelDecoration}
+          className={clsx(linkRecipe({ kind: children.kind }))}
```
federico-ercoles, line 115: Is this `clsx` needed any more?

packages/bento-design-system/src/Table/cells.tsx
```diff
@@ -85,17 +86,17 @@ export function LabelCell({ value, column: { align } }: CellProps<{}, LocalizedS
...
+  column: { align },
+}: CellProps<{}, { href: string; label: LocalizedString; isDisabled?: boolean }>) {
+  return (
+    <Box padding={16} textAlign={align}>
+      <Label size="large">
+        {makeTextChildrenFromElements(link(label, { href, disabled: isDisabled }))}
+      </Label>
+    </Box>
+  );
 }
```
federico-ercoles, line 100: Is it intended that `LinkCell` only supports real links and not link buttons?

federico-ercoles (commented):
Just a couple of questions, everything else looks ok to me.

packages/bento-design-system/src/util/TextChildren.tsx
veej: nope. Thanks 👍

packages/bento-design-system/src/Table/cells.tsx
veej: we already have an `IconButtonCell`. I don't think it makes a lot of sense to have a linkButton here

PR #142, opened 2022-04-13
Add defaultMenuSize to DropdownConfig

gabro:


packages/storybook/stories/index.tsx
```diff
@@ -119,6 +119,14 @@ export const {
   feedback: {
     background: FeedbackBackground,
   },
+  selectionControl: {
```
veej, line 122: are these related?

gabro: woops, no, commited junk by mistake...

veej (approved):

PR #145, opened 2022-04-15
Allow children in Link

gabro:
Two reasons:
- it's easier to plug into libraries like react-i18next
- it allows having a link "around" complex components


veej (approved):
💣

PR #146, opened 2022-04-15
Add more color options to Label and Body

gabro:


veej (approved):
looks 👍 even if I'm wondering if we should directly allow any text/foreground color instead.

gabro:
Good point, let's think about it, but maybe in a future PR

PR #150, opened 2022-04-19
#4820492: Slider should allow to drag with a different granularity

veej:
Closes [#4820492](https://buildo.kaiten.io/4820492)

### tests performed

### DIFF
--- a/packages/bento-design-system/src/SliderField/createSliderField.tsx
+++ b/packages/bento-design-system/src/SliderField/createSliderField.tsx
@@ -1,5 +1,6 @@
 import { useNumberFormatter } from "@react-aria/i18n";
 import { useSlider } from "@react-aria/slider";
+import { clamp } from "@react-aria/utils";
 import { useSliderState } from "@react-stately/slider";
 import { useRef, FunctionComponent } from "react";
 import { FieldType } from "../Field/createField";
@@ -17,6 +18,7 @@ type Props = (
   minValue: number;
   maxValue: number;
   step?: number;
+  dragStep?: number;
   autoFocus?: boolean;
 } & FormatProps;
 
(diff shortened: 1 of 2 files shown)


packages/bento-design-system/src/SliderField/createSliderField.tsx
```diff
@@ -42,11 +48,26 @@ export function createSliderField({
             onChange: (values: number[]) => props.onChange(values[0]),
           };
 
-    const state = useSliderState({
+    const _state = useSliderState({
```
gabro, line 51: super minor, ma possiamo usare un qualsiasi altro nome che non inizi con `_`? Usando `_` TS ignora i warning di unused, quindi se per caso diventa unused in un refactor non ce ne accorgiamo

packages/bento-design-system/src/SliderField/createSliderField.tsx
```diff
@@ -42,11 +48,26 @@ export function createSliderField({
...
     });
+    const state = props.dragStep
+      ? {
+          ..._state,
+          setThumbPercent: (index: number, percent: number) => {
+            const minValue = state.getThumbMinValue(index);
+            const maxValue = state.getThumbMaxValue(index);
+            const value = percent * (maxValue - minValue) + minValue;
+            const roundedValue =
+              roundToStep(roundToStep(value - minValue, props.dragStep!), props.step || 1) +
```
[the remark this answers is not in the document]
gabro: can you add this as a `NOTE` comment in the code?

gabro, line 64:
```suggestion
              roundToStep(roundToStep(value - minValue, props.dragStep!), props.step ?? 1) +
```

gabro: (unless you explicitly want to fallback to `1` in case the step is `0`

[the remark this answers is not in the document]
gabro: Also 🤯

gabro (changes requested):

packages/bento-design-system/src/SliderField/createSliderField.tsx
veej: I think we want to fallback to `1`, since it's the internal default if a step is not specified

gabro (approved):

PR #155, opened 2022-04-20
Improve storybook build

veej:


packages/bento-design-system/tsup.config.ts
```diff
@@ -2,7 +2,7 @@ import { defineConfig } from "tsup";
 import { vanillaExtractPlugin } from "@vanilla-extract/esbuild-plugin";
 
 export default defineConfig({
-  entry: ["src/index.ts", "src/vars.css.ts"],
+  entry: ["src/index.ts"],
```
veej, line 5: why are we building `vars.css.ts` as a separate file? In this way we're producing a `vars.css.js`, that will be processed also by the final project. The library is already producing a `.css` file with the vars, and in the final project we just need the TS vars, that can be re-exported from the index (so that they will be included in the `index.js` and will not be re-processed by the project's vanilla-extract plugin)

packages/storybook/.storybook/main.js
```diff
@@ -1,5 +1,27 @@
+const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+const { VanillaExtractPlugin } = require("@vanilla-extract/webpack-plugin");
+const path = require("path");
+
 module.exports = {
   stories: ["../stories/**/*.stories.*"],
   addons: ["@storybook/addon-links", "@storybook/addon-essentials", "storybook-addon-themes"],
   framework: "@storybook/react",
+  core: {
+    builder: "webpack5",
```
veej, line 10: since I was there, I configured storybook to use the new webpack5

packages/storybook/.storybook/main.js
```diff
@@ -1,5 +1,27 @@
...
+  },
+  webpackFinal: (baseConfig) => {
+    const { module = {}, plugins = {} } = baseConfig;
+
+    // by default, the rules for jsx|tsx files are including the whole monorepo root,
+    // but it's enough to check only the storybook package
+    const tsxRules = module.rules.filter((rule) => rule?.test?.test("test.tsx"));
```
veej, line 17: Not sure if this was the cause for the hang during the storybook-start, but there's no need to include the whole monorepo root. We just want to transpile the files in the storybook package

veej: re-tested without these lines and the issue reappears. I suppose storybook is trying to react both to the change in the library itself, and at the same time to the change in the node_modules after the library is built by tsup

gabro (approved):

PR #156, opened 2022-04-21
Default theme

veej:
Tested in a brand-new CRA application, with just this `index.ts`:

```ts
import "@buildo/bento-design-system/lib/index.css"
import { bentoSprinkles } from '@buildo/bento-design-system/lib/defaultTheme'
import "@buildo/bento-design-system/lib/defaultTheme.css"
import { createBentoComponents } from '@buildo/bento-design-system'

const { AreaLoader, Inset } = createBentoComponents(bentoSprinkles)

function App() {
  return (
    <AreaLoader />
  );
}

export default App;
```

image


packages/bento-design-system/src/defaultTheme.ts
```diff
@@ -0,0 +1 @@
+import "./defaultTheme.css";
```
veej, line 1: didn't find a better way to generate a `defaultTheme.css`.
if we include `defaultTheme.css.ts` directly as an entry in the tsup config, we get a `defaultTheme.css.css`

gabro (approved):

PR #160, opened 2022-04-22
More flexible createBentoComponents

gabro:
This PR allows `createBentoComponents` to accept:

- 0 parameter (uses `bentoSprinkles` and `defaultConfigs`)
- only sprinkles
- only config
- both

In order to preserve the type inference all the possible cases are implemented via overloads.

The tricky part is that we need to specify a return type for the overloads, but we want that to be inferred from the implementation (listing all the components in the return type is a bit tedious...). <s>Unfortunately the current version of TS doesn't allow invoking `ReturnType` on a generic function, but the upcoming version (4.7) will allow it!

This PR currently depends on TypeScript 4.7.0-beta, and I've tested it works there, but I'd probably wait for TS 4.7.0 stable to be out before merging it.</s>

We now can merge this since we upgrade to 4.7.0!

### DIFF
--- a/packages/bento-design-system/src/createBentoComponents.ts
+++ b/packages/bento-design-system/src/createBentoComponents.ts
@@ -303,5 +303,7 @@ function internalCreateBentoComponents<
   };
 }
 
-type R<SprinklesFn extends typeof bentoSprinkles, CustomChipColor extends string = never>
-   = ReturnType<typeof internalCreateBentoComponents<SprinklesFn, CustomChipColor>>;
+type R<
+  SprinklesFn extends typeof bentoSprinkles,
+  CustomChipColor extends string = never
+> = ReturnType<typeof internalCreateBentoComponents<SprinklesFn, CustomChipColor>>;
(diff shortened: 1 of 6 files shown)


packages/bento-design-system/src/createBentoComponents.ts
```diff
@@ -254,3 +283,6 @@ export function createBentoComponents<
     useComponentsShowcase,
   };
 }
+
+type R<SprinklesFn extends typeof bentoSprinkles, CustomChipColor extends string = never>
+   = ReturnType<typeof internalCreateBentoComponents<SprinklesFn, CustomChipColor>>;
```
gabro, line 288: this is the new thing in 4.7, i.e. the fact we can do `typeof foo<A, B>` where previously that would have been a parser error.

In fact, `prettier` will also choke on it since the parser hasn't been updated to suppor that syntax yet (it's called "instantiation expressions", see https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#instantiation-expressions)

veej:
@gabro there are conflicts to resolve.

gabro:
@veej done now!

veej (approved):
💯

PR #165, opened 2022-04-24
Make LocalizedString type configurable via module augmentation

gabro:
This PR makes `LocalizedString` configurable via module augmentation.

`LocalizedString` now defaults to `string`, which makes it a bit more friendly when approaching the library, since it works out of the box with any existing localization library.

Still, clients can override the type to something stricter, like the branded string we were previously using. The override looks like:

```ts
// my-project/app/src/bento.d.ts
import "@buildo/bento-design-system";

declare module "@buildo/bento-design-system" {
  interface TypeOverrides {
    LocalizedString: StrictLocalizedString
}
```

This is breaking because the type of `LocalizedString` is now less restrictive, but it won’t actually cause any compilation error downstream.


packages/bento-design-system/src/util/LocalizedString.ts
```diff
@@ -1,9 +1,12 @@
...
-export type LocalizedString = string & {
-  readonly LocalizedString: "LocalizedString";
-};
+interface ConfigurableTypes {
+  LocalizedString: string;
+}
+
+export interface TypeOverrides {}
+
+type ResolvedTypes = Omit<ConfigurableTypes, keyof TypeOverrides> & TypeOverrides;
```
gabro, line 7: this is a trick I borrowed from react-i18next: we compute the final type as the merge of two objects. The second one (`TypeOverrides`) takes precedence and it’s initially empty.

Users of the library can then use module augmentation to inject their own types and they will “win” over the default one.

packages/bento-design-system/test/tsconfig.json
```diff
@@ -4,6 +4,11 @@
   "compilerOptions": {
     "jsx": "react-jsx",
     "noEmit": true,
-    "preserveSymlinks": true
+    "preserveSymlinks": true,
+    "baseUrl": ".",
+    "paths": {
+      "@buildo/bento-design-system": ["../src"]
```
gabro, line 10: this is needed for module augmentation to work using this alias in the tests

packages/bento-design-system/src/util/ConfigurableTypes.ts
```diff
@@ -0,0 +1,13 @@
+import { Object } from "ts-toolbelt";
+
+interface ConfigurableTypes {
+  LocalizedString: string;
```
giogonzo, line 4: we could also include `number` here while at it?
```suggestion
  LocalizedString: string | number;
```

gabro: done!

packages/bento-design-system/src/util/LocalizedString.ts
```diff
@@ -1,9 +1,8 @@
-/**
- * A string that has been localized
- */
-export type LocalizedString = string & {
-  readonly LocalizedString: "LocalizedString";
-};
+import { ConfiguredTypes } from "./ConfigurableTypes";
+
+export type LocalizedString = ConfiguredTypes["LocalizedString"];
```
gabro, line 3: Note to self, maybe it's better to do something like `(string | number) & ConfiguredTypes["LocalizedString"]` so that we have a baseline to work with. Otherwise a client can potentially override this with a unrelated type and then the whole thing becomes unsafe

packages/website/docs/Advanced/strict-localization.md
```diff
@@ -0,0 +1,134 @@
...
+        onPress={() => {}}
+        kind="solid"
+        hierarchy="primary"
+      />
+    </>
+  );
+}
+```
+
+Great! Now all Bento components will complain if we accidentally forget to localized a string that must be presented to the user 🎉
```
gabro, line 69:
```suggestion
Great! Now all Bento components will complain if we accidentally forget to localize a string that must be presented to the user 🎉
```

veej (approved):
Looks great 💯

PR #166, opened 2022-04-26
add component Props doc

veej:


gabro:
@veej can you rebase on main so that we can test if the preview works?

gabro (approved):
Nice!

PR #169, opened 2022-04-26
Docs components

veej:
This PR just adds an .mdx file for every Bento component, without caring too much about the content. I didn't implement examples or checked the generated configs/props yet. I just checked the links to github/storybook/zeroheight.

We have an issue with create functions taking dependency components, which are not recognized as react components, thus ignored by the docgen. Also, for at least one component, the create function doesn't have a config, so the docgen is taking the list of dependency components as config keys.

### DIFF
--- a/packages/website/src/components/PropTable.tsx
+++ b/packages/website/src/components/PropTable.tsx
@@ -2,7 +2,6 @@ import * as React from "react";
 import useGlobalData from "@docusaurus/useGlobalData";
 
 export const PropTable = ({ name }) => {
-  console.log(useGlobalData()["docusaurus-plugin-react-docgen-typescript"].default as any);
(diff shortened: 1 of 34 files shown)


packages/website/src/components/PropTable.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from "react";
 import useGlobalData from "@docusaurus/useGlobalData";
 
 export const PropTable = ({ name }) => {
+  console.log(useGlobalData()["docusaurus-plugin-react-docgen-typescript"].default as any);
```
gabro, line 5: leftover?

gabro: by the way you can inspect global data in the `.docusaurus` folder

gabro (approved):

PR #172, opened 2022-04-27
reorder docs pages + minor fixes

veej:
not sure what to do with the "Atomic Styles" and "Vanilla Extract" pages.
IMHO "Atomic styles" should appear early enough in the list, as an introduction to the general styling strategy. Not sure if it's better to move it in the `Customization` section though.
"Vanilla Extract", instead, is kinda collateral and linked by other resources. Left as the last element atm


packages/website/docs/03-Customization/03-atoms-augmentation.md
```diff
@@ -0,0 +1,116 @@
...
+
+// our new custom variables
+export const customVars = createGlobalTheme(":root", {
+  color: {
+    specialBackgroundColor: "#3C6FD6",
+  },
+});
+```
+
+Then, we must tell Bento this new token can be used as a value for the `background` CSS property, by extending Bento's `statusProperties`:
```
veej, line 54: I replaced `color` with `background` in various places in this file, to align the description to the new example

gabro (approved):

PR #173, opened 2022-04-27
#4887422: Table column hints shall support onClick

veej:
Closes [#4887422](https://buildo.kaiten.io/4887422)

### tests performed


Phyele:
Tooltip trigger is a 12px icon "Info"

gabro (approved):

Phyele (approved):

PR #176, opened 2022-04-28
remove config table, expand enum types

veej:


gabro (approved):
It's already a good improvement so 👌 let's merge it

PR #177, opened 2022-04-28
Docs components examples

veej:


packages/website/src/snippets/Components/Actions.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from "react";
+import { Actions } from "..";
+import { formatMessage } from "../formatMessage";
+
+export default function ActionsExample() {
+  return (
+    <Actions
+      size="medium"
+      primaryAction={{
+        label: formatMessage("Ok"),
```
gabro, line 10: No need for formatMessage anymore, I would probably remove it in the examples

gabro (changes requested):
💪🎉 just one comment

packages/website/src/snippets/Components/Actions.tsx
veej: I was unsure, but maybe you're right... better to keep it simple.

gabro (approved):

PR #180, opened 2022-04-29
#3129679: Allow to configure the selected locale in DesignSystemProvider

veej:
Closes [#3129679](https://buildo.kaiten.io/3129679)

### tests performed

Tested adding `locale="it"` in the `DesignSystemProvider` used for storybook, and the `NumberField` turned from accepting dot-separated decimals to comma-separated decimals. 

It should automatically work also for `DateField`, since we're using `useDateFormatter`


gabro (approved):
👌

PR #184, opened 2022-05-03
fix datepicker

veej:
Should solve the issue with the DateField throwing an exception when opened in docs.
Not sure exactly why it works, but I tried to build and serve the docs locally:

```
pnpm website-build
npx serve packages/website/build
```

and it seems to work fine.


veej:
Seems to work fine also in the deployed docs 🤷‍♂️

gabro (approved):
🤷‍♂️

PR #190, opened 2022-05-04
#3087357: Chip - Fix

veej:
Closes [#3087357](https://buildo.kaiten.io/3087357)

### tests performed

> A Test Plan is used to show what you tested to make sure your code works fine. You should write here all that you did to test, and provide some results of your testing.

> These results can be screenshots, query results, or even just pointers to unit tests that successfully passed.

### tests not performed (domain coverage)

> At times not everything can be tested, and writing what hasn't been tested is just as important as writing what has been tested.

> An example of partial test is a field displaying 4 possible values. If 3 values are tested, with screenshots, and 1 is not, then it should be mentioned here.

PR #191, opened 2022-05-04
Review of #190 

gabro:
@veej an alternative to the solution proposed in #190


veej (approved):
💯

veej (changes requested):
@gabro looking at the screenshot, I think there's something wrong in the chip height... 
This is my solution:
image

And here how it appears with yours:
image

veej:
🤔 I tried to open it in storybook and I see this:
image

gabro:
woops, fixed now

veej (approved):
Looks great 💯

PR #195, opened 2022-05-04
#4862669: Banner, Modal: separate padding config into paddingX/paddingY

veej:
Closes [#4862669](https://buildo.kaiten.io/4862669)

### tests performed


gabro:
Oppure no, ma tagghiamola come breaking

veej:
Taggata come breaking

gabro (approved):

PR #204, opened 2022-05-06
Support stateful components and strip TypeScript syntax in Playroom

gabro:
Two main improvements:
- strip TypeScript syntax from the examples (this would trip up Playroom which expects JavaScript)
- carry over the `React.useState` uses (while still hiding them from the snippets on the website) so that stateful components work in Playroom

As examples, check out `DateField` (which has both a TypeScript annotation and it uses state) and see how it gets translated when jumping to Playroom


veej (approved):
Looks 💯

PR #211, opened 2022-05-09
Add rightAccessory to Tabs tab

gabro:


packages/storybook/stories/index.tsx
```diff
@@ -123,4 +123,5 @@ export const {
   },
 });
 
+export const FolderTabs = Tabs;
```
gabro, line 126: so that Playroom "sees" a `Tabs` element

packages/bento-design-system/src/Tabs/createTabs.tsx
```diff
@@ -60,9 +71,12 @@ export function createTabs(config: TabsConfig) {
           {icon && (
             <Column width="content">{icon({ size: config.iconSize, color: "inherit" })}</Column>
           )}
-          <Label size={config.labelSize[size]} uppercase={config.uppercaseLabel}>
-            {label}
-          </Label>
+          <Column width="content">
```
veej, line 74: having `width="content"` on the label or on the rightAccessory doesn't make any difference in this case, right?

veej (approved):

PR #220, opened 2022-05-27
allow to pass complex Children to StepperStep label

veej:


packages/bento-design-system/src/Stepper/createStepper.tsx
```diff
@@ -23,21 +24,13 @@ export function createStepper(config: StepperConfig) {
         {steps.map(({ label }, index) => {
           const status =
             index < currentStep ? "done" : index === currentStep ? "inProgress" : "todo";
-          return <Step label={label} index={index} status={status} key={label} />;
+          return <Step label={label} index={index} status={status} key={index} />;
```
gabro, line 27: I've done a similar refactor for ListItem, I would suggest to include an optional `key` property to StepProps so that once can customize it in those cases where using index is problematic

gabro (changes requested):
One tiny suggestion

packages/bento-design-system/src/Stepper/createStepper.tsx
veej: makes sense

gabro (approved):

PR #226, opened 2022-05-31
Use border instead of outline for Card, Menu, and Toast

gabro:
This fixes the behavior on Safari.

I was tempted to also remove `outline...` from the atoms, but it would be a breaking change, so I'll save that for a future PR.

Before:
image

After:
image


veej (approved):
💯

PR #227, opened 2022-05-31
Add explicit grid-auto-rows: max-content to Table rows

gabro:
This fixes a bug in Safari which seems to manifest in these
circumstances:
- the Table is included in a container that grows vertically (e.g. a
  Stack)
- the Table scrolls horizontally (i.e. some columns overflow)

In this scenario, all rows become taller (over ~150px).
This may be due to how Safari interprets height: 100% (since wrapping
the Table in a simple `Box` seems to fix it) but it's only an educated
guess at this point.


veej (approved):
💯

PR #235, opened 2022-06-16
Add selectAll / clearAll to SelectField

gabro:


gabro:
Right, that's fair, I'll add the label

PR #244, opened 2022-06-27
Modal: add 'kind', deprecate 'isDestructive', add 'size="wide"'

gabro:


packages/bento-design-system/src/Modal/createModal.tsx
```diff
@@ -11,19 +11,23 @@ import { useDefaultMessages } from "../util/useDefaultMessages";
...
+export type ModalKind = "normal" | "warning" | "destructive";
 type Props = {
   title: LocalizedString;
   children: Children;
   primaryAction?: Omit<ButtonProps, "kind" | "hierarchy" | "size">;
   secondaryAction?: Omit<ButtonProps, "kind" | "hierarchy" | "size">;
   onClose: () => void;
   closeButtonLabel?: LocalizedString;
+  /** @deprecated use kind='destructive' instead */
   isDestructive?: boolean;
```
gabro, line 26: not removing this to avoid breaking changes

packages/bento-design-system/src/Modal/createModal.tsx
```diff
@@ -86,11 +90,24 @@ export function createModal(
 
     const { defaultMessages } = useDefaultMessages();
 
+    const kind = props.kind == null && props.isDestructive ? "destructive" : props.kind ?? "normal";
```
gabro, line 93: The behavior is to preserve backward compatibility:
- if the user is using `isDestructive` we just look at that
- if the user sets `kind` then we ignore `isDestructive`

Phyele (approved):

Phyele (approved):

PR #249, opened 2022-07-06
fix print of Table with empty headers

veej:
## Test plan

- Remove an headerLabel from a column in the Table story
- Render the Table story in isolation: http://localhost:6006/iframe.html?id=components-table--simple&args=&viewMode=story
- Try to print the page with the "print" browser functionality

### before
=> print loop. Print preview is not generated

### after
=> the preview is correctly generated in a few milliseconds


gabro (approved):
Ok as a quick fix, but it would be nice to understand why so that we can also check other components

PR #254, opened 2022-07-10
Fix borderStyle atoms

gabro:
Remove the duplicate `borderStyle` atoms from `unconditionalProperties`.

I’ve also moved the `borderBottomStyle` to `statusProperties` for consistency


veej (approved):
💯

PR #265, opened 2022-07-13
Update to React 18 and related dependencies

gabro:
We also keep React 17 compatibility and add React 18 to the peer dependencies


packages/bento-design-system/src/Table/types.ts
```diff
@@ -25,7 +26,7 @@ declare module "react-table" {
 
 export type Column<A extends string, D extends object, V> = {
   accessor: A;
-  Cell: (props: CellProps<D, V>) => Children;
+  Cell: (props: CellProps<D, V>) => ReturnType<FunctionComponent>;
```
gabro, line 29: This is due to a slight change in the definition of `FunctionComponent`, which is ultimately used by the `Renderer` type in react-table.

It's technically breaking at the type-level, since we don't allow `undefined` to be returned anymore by the `Cell` function, but I think we can ignore this in practice.

packages/bento-design-system/src/react-input-mask.d.ts
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import "react-input-mask";
+
+declare module "react-input-mask" {
+  interface Props {
+    children: (inputProps: InputHTMLAttributes<HTMLInputElement>) => JSX.Element;
```
veej, line 6: should be optional?

packages/bento-design-system/src/react-input-mask.d.ts
```diff
@@ -0,0 +1,8 @@
...
+  }
+}
```
gabro, line 8: needed since React 18 removed the "implicit" `children` prop, and `react-input-mask` types have not been updated in a while.

veej (commented):
minor doubt + tests are failing in the CI

packages/bento-design-system/src/react-input-mask.d.ts
gabro: we use it only in one case, not sure it matters, but I can improve the type, yes

gabro:
@veej fixed the tests (it was a breaking change in `@testing-library/user-event`)

veej (approved):
💯

PR #273, opened 2022-07-14
Add TextArea component

Reggionick:


packages/bento-design-system/src/TextArea/createTextArea.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { inputRecipe } from "../Field/Field.css";
+import { FieldProps } from "../Field/FieldProps";
+import { bodyRecipe } from "../Typography/Body/Body.css";
+import { FieldType } from "../Field/createField";
+import { InputConfig } from "../Field/Config";
+
+type Props = FieldProps<string> & {
+  placeholder: LocalizedString;
+  isReadOnly?: boolean;
+};
```
gabro, line 14: can you also include the standard `rows` attribute to pass down to the textarea, so that a user can control the number of rows?

gabro (changes requested):
Nice job! A few comments inline

gabro (approved):
:shipit:

gabro:
@Reggionick the readonly version looks strange (I've marked as denied on Chromatic). It seems it does not have the correct background, can you check?

PR #278, opened 2022-07-15
Added searchable prop to SelectField

Reggionick:


Reggionick:
> How does it work with the multi-select? I think in case `multi={true}` it was already searchable? If that's the case, maybe we can limit the `searchable` prop to the `isMulti={false}` case

Before this changes `react-select` was enabling search by default. With multi enabled it still working, I would leave the prop as it's right now. What do you think?

packages/bento-design-system/src/SelectField/createSelectField.tsx
```diff
@@ -174,7 +176,7 @@ export function createSelectField(
           hideSelectedOptions={false}
           menuSize={menuSize}
           menuIsOpen={isReadOnly ? false : undefined}
-          isSearchable={isReadOnly ? false : undefined}
+          isSearchable={isReadOnly ? false : searchable ?? false}
```
gabro, line 179:
```suggestion
          isSearchable={isReadOnly ? false : searchable ?? true}
```

We discussed this internally, and let's keep it `true` by default so to avoid a breaking change

gabro (changes requested):

PR #282, opened 2022-07-15
#281: Sticky table header and fixed table height

### LINKED ISSUE #281: Sticky table header and fixed table height
Add an option to make Table headers sticky and also add the ability to restrict the Table height to a fixed value (so that the Table can scroll and the header can effectively be sticky).

gabro:
Closes #281


federico-ercoles (approved):
LGTM 👍🏻

PR #285, opened 2022-07-19
Added show password to password field

Reggionick:


packages/bento-design-system/src/Icons/IconEye.tsx
```diff
@@ -4,7 +4,18 @@ import { svgIconProps } from "./svgIconProps";
...
+      <path
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        d="M12 9.82982C11.3962 9.82982 10.8171 10.0697 10.3902 10.4966C9.96324 10.9236 9.72339 11.5026 9.72339 12.1064C9.72339 12.7102 9.96324 13.2893 10.3902 13.7162C10.8171 14.1432 11.3962 14.383 12 14.383C12.6038 14.383 13.1829 14.1432 13.6098 13.7162C14.0367 13.2893 14.2766 12.7102 14.2766 12.1064C14.2766 11.5026 14.0367 10.9236 13.6098 10.4966C13.1829 10.0697 12.6038 9.82982 12 9.82982ZM8.78038 8.88682C9.63428 8.03293 10.7924 7.55322 12 7.55322C13.2076 7.55322 14.3657 8.03293 15.2196 8.88682C16.0735 9.74071 16.5532 10.8988 16.5532 12.1064C16.5532 13.314 16.0735 14.4721 15.2196 15.326C14.3657 16.1799 13.2076 16.6596 12 16.6596C10.7924 16.6596 9.63428 16.1799 8.78038 15.326C7.92649 14.4721 7.44678 13.314 7.44678 12.1064C7.44678 10.8988 7.92649 9.74071 8.78038 8.88682Z"
+        fill="#525E6F"
```
gabro, line 11: as discussed on Slack, let's remove this fill so that it gets inherited from the svg props

packages/bento-design-system/src/Icons/IconEye.tsx
```diff
@@ -4,7 +4,18 @@ import { svgIconProps } from "./svgIconProps";
...
+      />
+      <path
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        d="M2.33808 12.1064C3.74446 16.0864 7.54152 18.9363 12 18.9363C16.4595 18.9363 20.2556 16.0865 21.6619 12.1064C20.2556 8.12639 16.4595 5.27661 12 5.27661C7.54152 5.27661 3.74446 8.12645 2.33808 12.1064ZM0.0522882 11.7654C1.64722 6.68641 6.39245 3 12 3C17.6088 3 22.3528 6.68647 23.9477 11.7654C24.0174 11.9874 24.0174 12.2255 23.9477 12.4475C22.3528 17.5264 17.6088 21.2129 12 21.2129C6.39245 21.2129 1.64722 17.5265 0.0522882 12.4475C-0.0174294 12.2255 -0.0174294 11.9874 0.0522882 11.7654Z"
+        fill="#525E6F"
```
gabro, line 17: same

packages/bento-design-system/src/Icons/IconEyeClosed.tsx
```diff
@@ -0,0 +1,15 @@
...
+        // fill="#525E6F"
```
gabro, line 11: same

packages/bento-design-system/src/TextField/createTextField.tsx
```diff
@@ -31,34 +38,71 @@ export function createTextField(config: InputConfig, { Field }: { Field: FieldTy
...
+    const iconProps: IconProps = useMemo(() => ({ size: config.passwordIconSize || 24 }), []);
+
+    const type = useMemo(() => {
+      if (props.type === "password") {
+        return showPassword ? "text" : "password";
+      } else {
+        return props.type || "text";
+      }
+    }, [props.type, showPassword]);
+
```
gabro, line 58: I would remove all the `useMemo` uses here: they make the code a bit harder to read, and I don't think there is any tangible benefit (the computations a very cheap)

packages/bento-design-system/src/TextField/createTextField.tsx
```diff
@@ -31,34 +38,71 @@ export function createTextField(config: InputConfig, { Field }: { Field: FieldTy
...
+          : config.passwordShowIcon ?? IconEyeClosed,
+      [showPassword]
+    );
+    const iconProps: IconProps = useMemo(() => ({ size: config.passwordIconSize || 24 }), []);
+
+    const type = useMemo(() => {
+      if (props.type === "password") {
+        return showPassword ? "text" : "password";
+      } else {
+        return props.type || "text";
```
gabro, line 55:
```suggestion
        return props.type ?? "text";
```

gabro (changes requested):

packages/bento-design-system/src/TextField/createTextField.tsx
```diff
@@ -31,34 +38,63 @@ export function createTextField(config: InputConfig, { Field }: { Field: FieldTy
       inputRef
     );
 
+    const [showPassword, setShowPassword] = useState(false);
+    const passwordIcon = showPassword
+      ? config.passwordHideIcon ?? IconEye
+      : config.passwordShowIcon ?? IconEyeClosed;
+
+    const iconProps: IconProps = { size: config.passwordIconSize || 24 };
```
gabro, line 46:
```suggestion
    const iconProps: IconProps = { size: config.passwordIconSize ?? 24 };
```

packages/bento-design-system/src/TextField/createTextField.tsx
```diff
@@ -31,34 +38,63 @@ export function createTextField(config: InputConfig, { Field }: { Field: FieldTy
...
+
+    const type =
+      props.type === "password" ? (showPassword ? "text" : "password") : props.type ?? "text";
```
gabro, line 49:
```suggestion
      (props.type === "password" && !showPassword) ? "password" : "text";
```

gabro (changes requested):

packages/bento-design-system/src/TextField/createTextField.tsx
```diff
@@ -31,34 +38,62 @@ export function createTextField(config: InputConfig, { Field }: { Field: FieldTy
...
+              justifyContent="center"
+              alignItems="center"
+              paddingX={16}
+              top={0}
+              bottom={0}
+              right={0}
+              onClick={() => setShowPassword((prevValue) => !prevValue)}
+            >
+              {passwordIcon(iconProps)}
+            </Box>
```
gabro, line 94: Can you use `IconButton` here? `IconButton` gives you a few features such as bigger target area, proper hover, keyboard management and accessibility via an aria-label.

You probably still need the `Box` for positioning the element, but you can remove the `onClick` from it

packages/bento-design-system/src/TextField/createTextField.tsx
```diff
@@ -31,34 +38,62 @@ export function createTextField(config: InputConfig, { Field }: { Field: FieldTy
       inputRef
     );
 
+    const [showPassword, setShowPassword] = useState(false);
+    const passwordIcon = showPassword
+      ? config.passwordHideIcon ?? IconEye
+      : config.passwordShowIcon ?? IconEyeClosed;
+
```
gabro, line 45: The two icons are swapped, it should show `IconEye` when the password is concealed and `IconEyeClosed` when is in clear text

gabro (changes requested):

packages/bento-design-system/src/TextField/createTextField.tsx
Reggionick: A label is required for IconButton. What should be set?

gabro: right, those should come from default messages and be customizable. You can observe the same pattern in `Chip`, see how the `dismissButtonLabel` is handled (optional prop on the component, mandatory in `DefaultMessages`).

This will also require to update the `defaultMessages` in a few spots (you can just use `"Hide password"` / `"Show password"`)

packages/bento-design-system/src/TextField/createTextField.tsx
```diff
@@ -31,34 +45,69 @@ export function createTextField(config: InputConfig, { Field }: { Field: FieldTy
...
+              display="flex"
+              justifyContent="center"
+              alignItems="center"
+              paddingX={16}
+              top={0}
+              bottom={0}
+              right={0}
+            >
+              <IconButton
+                size={config.passwordIconSize ?? 24}
```
gabro, line 101: The default should not live here, but in the `defaultConfig`

packages/bento-design-system/src/Field/Config.ts
```diff
@@ -27,6 +27,9 @@ export type InputConfig = {
   radius: BentoSprinkles["borderRadius"];
   fontSize: BodyProps["size"];
   iconSize: IconProps["size"];
+  passwordShowIcon?: (props: IconProps) => Children;
+  passwordHideIcon?: (props: IconProps) => Children;
+  passwordIconSize?: IconProps["size"];
```
gabro, line 32: the config entries should always be required (we process them to make them all optional when exposing this to the users)

gabro (changes requested):
A couple of more comments, plus the CI is failing

packages/bento-design-system/src/TextField/createTextField.tsx
Reggionick: Also `passwordHideIcon` and `passwordShowIcon`?

gabro: yes, actually, good point

gabro:
CI is failing

gabro (approved):

PR #287, opened 2022-07-19
Added Avatar image prop

Reggionick:


packages/bento-design-system/src/Avatar/createAvatar.tsx
```diff
@@ -35,7 +50,15 @@ export function createAvatar(config: AvatarConfig) {
...
         >
-          {initial ? (
+          {imageSrc && imageAvailable !== false ? (
+            <img
+              style={{ display: imageAvailable ? "block" : "none" }}
+              src={imageSrc}
+              className={avatarImage({})}
+              onLoad={handleImageLoaded}
+              onError={handleImageErrored}
+            />
```
gabro, line 60: I propose we use `<Box as="img">` here, which also gives you the ability to use `display` directly (`<Box display={imageAvailable ? "block" : "none"} />`)

packages/bento-design-system/src/Avatar/Avatar.css.ts
```diff
@@ -23,3 +23,9 @@ export const avatarRecipe = strictRecipe({
     },
   },
 });
+
+export const avatarImage = strictRecipe({
+  base: bentoSprinkles({
+    width: "full",
+  }),
+});
```
gabro, line 31: no need to use a recipe here, you can directly use `bentoSprinkles`.

gabro (changes requested):
A couple of comments inline, plus:

- can we add a story for the "fallback" case? (e.g. with a broken url which doesn't load)
- I think we need to clip the image (e.g. try with a squared image, I would expect it not to overflow)

PR #313, opened 2022-07-26
Improve table sorting example

marcopiii:


packages/bento-design-system/src/Table/createTable.tsx
```diff
@@ -71,9 +71,22 @@ type SortFn<C extends ReadonlyArray<ColumnType<string, {}, any>>> = (
...
 type SortingProps<C extends ReadonlyArray<ColumnType<string, {}, any>>> =
   | {
-      /** This function must be memoized to avoid infinite re-renderings */
+      /**
+       * `customSorting` is meant to be used to provide a custom sorting logic
+       * to the table. It supports cross-columns comparison.
+       *
+       * The sorting is still performed by the table.
+       *
+       * This function must be memoized to avoid infinite re-renderings */
```
gabro, line 80:
```suggestion
      /**
       * `customSorting` can be used to customize the sorting logic of the table. It supports cross-columns comparison.
       *
       * The sorting is still performed by the table, in an uncontrolled fashion.
       *
       * This function must be memoized to avoid infinite re-renderings */
```

packages/bento-design-system/src/Table/createTable.tsx
```diff
@@ -71,9 +71,22 @@ type SortFn<C extends ReadonlyArray<ColumnType<string, {}, any>>> = (
...
-      /** This function must be memoized to avoid infinite re-renderings */
+      /**
+       * `onSort` is meant to be used when the sorting and/or filtering is performed
+       * outside the table (for example by a backend service).
+       *
+       * If `onSort` is provided, the table expects to receive all and only the rows
+       * to display as the `data` property and it will **not** apply any sorting on them.
+       *
+       * This function must be memoized to avoid infinite re-renderings */
       onSort?: (sortBy: Array<SortingRule<C>>) => void;
```
gabro, line 90:
```suggestion
      /**
       * `onSort` can be used to implement sorting in a controlled fashion. It's typical use case is when sorting and/or filtering is performed externally, for example by a backend service.
       *
       * If `onSort` is provided the table won't perform any sorting on the rows given in input via the `data` property, as the caller is now in charge of performing the sorting.
       *
       * This function must be memoized to avoid infinite re-renderings */
      onSort?: (sortBy: Array<SortingRule<C>>) => void;
```

gabro (approved):
👍 approved, I have a couple of non-blocking suggestions on the documentation

gabro (approved):

PR #329, opened 2022-07-28
#292: Autofocus on fields

### LINKED ISSUE #292: Autofocus on fields
### Discussed in https://github.com/buildo/bento-design-system/discussions/286

<div type='discussions-op-text'>
<sup>Originally posted by **mperrando** July 19, 2022</sup>
</div>

As discussed, it would be convenient for fields to expose an `autoFocus` prop. I would add it to:
- all form fields (text fields, number field, radio field, etc)
- SearchBar
- Form (with the same semantics as Modal, i.e. autofocus the first focusable children)

Implementation notes:
- for fields and searchbar, we can probably pass down the `autoFocus` prop to the native browser component
- for `Form` I would use the same approach as `Modal` and wrap the children in a `<FocusScope autoFocus={props.autoFocus}>`

marcopiii:
Closes #292

### tests performed

- By using the arg `autoFocus: true` in the stories, the components `TextField`, `TextArea`, `SelectField`, `RadioGroupField`, `SearchBar`, `CheckboxGroupField`, `Slider`, `Form` and `DateField` actually get the focus.

### DIFF
--- a/packages/bento-design-system/src/Form/Form.tsx
+++ b/packages/bento-design-system/src/Form/Form.tsx
@@ -44,7 +44,7 @@ export function createForm(
     autoFocus,
   }: Props) {
     return (
-      <FocusScope contain restoreFocus autoFocus={autoFocus}>
+      <FocusScope autoFocus={autoFocus}>
         <Box as="form" onSubmit={(e) => e.preventDefault()}>
           <ContentBlock maxWidth={700}>
             <Stack space={config.formSpacing}>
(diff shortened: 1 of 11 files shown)


packages/bento-design-system/src/Form/Form.tsx
```diff
@@ -39,34 +41,37 @@ export function createForm(
...
-                size={actionsSize}
-                primaryAction={submitButton}
-                secondaryAction={secondaryButton}
-                error={error}
-              />
-            )}
-          </Stack>
-        </ContentBlock>
-      </Box>
+      <FocusScope contain restoreFocus autoFocus={autoFocus}>
```
gabro, line 47:
```suggestion
      <FocusScope autoFocus={autoFocus}>
```

We only want `autoFocus` here, we don't need `contain` and `restoreFocus` since this is not a `Modal`

gabro (changes requested):
One comment inline, plus I would also add it to `NumberInput`

gabro (approved):

PR #330, opened 2022-07-28
#319: Use `{ custom: number }` convention for Illustration size

### LINKED ISSUE #319: Use `{ custom: number }` convention for Illustration size
`Illustration` accepts a set of fixed sizes and also a generic `number` for custom sizes.

That was one of the first times we allowed arbitrary values in the DS, and since then we standardized on using `{ custom: number }` in union with the known values (this leads to better type inference, and it introduces an appropriate amount of friction when using custom values)

This means that instead of:

```tsx
  size: 24 | 32 | 40 | 80 | 160 | number;
```

we should have

```tsx
  size: 24 | 32 | 40 | 80 | 160 | { custom: number };
```

instead. 

This is unfortunately a breaking change, but I think it's worth it.

marcopiii:
Closes #319


veej (approved):
Looks 💯

PR #340, opened 2022-08-02
#320: Use official floating-ui dependency instead of fork

### LINKED ISSUE #320: Use official floating-ui dependency instead of fork
Back in https://github.com/buildo/bento-design-system/pull/39 we started using a fork of floating ui due to this issue: https://github.com/floating-ui/floating-ui/pull/1529

Since then, Floating UI release its version 1.0.0 which should have fixed this issue in https://github.com/floating-ui/floating-ui/pull/1678

We should try going back to the main dependency instead of the fork we're using now

marcopiii:
Closes #320

### tests performed

- It builds, it runs, everything seems unchanged

### DIFF
--- a/packages/bento-design-system/package.json
+++ b/packages/bento-design-system/package.json
@@ -39,9 +39,9 @@
   "dependencies": {
     "@datepicker-react/hooks": "^2.8.4",
     "@dessert-box/react": "^0.4.0",
-    "@floating-ui/core": "^1.0.0",
-    "@floating-ui/dom": "^1.0.0",
-    "@floating-ui/react-dom": "^1.0.0",
+    "@floating-ui/core": "1.0.0",
+    "@floating-ui/dom": "1.0.0",
+    "@floating-ui/react-dom": "1.0.0",
     "@fontsource/ibm-plex-sans": "^4.5.11",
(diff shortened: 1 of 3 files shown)


packages/bento-design-system/package.json
```diff
@@ -36,15 +36,12 @@
...
-  "dependenciesComments": {
-    "@farzadsh/floating-ui-*": "NOTE(gabro): we are using a fork of Floating UI due to an issue with the build of the published package. See https://github.com/floating-ui/floating-ui/pull/1529"
-  },
   "dependencies": {
     "@datepicker-react/hooks": "^2.8.4",
     "@dessert-box/react": "^0.4.0",
-    "@farzadsh/floating-ui-core": "^0.4.1",
-    "@farzadsh/floating-ui-dom": "^0.2.1",
-    "@farzadsh/floating-ui-react-dom": "^0.4.4",
+    "@floating-ui/core": "^1.0.0",
```
gabro, line 42: While we’re at it, let’s also pin the dependencies (so, no ^)

gabro (changes requested):
One tiny change

gabro (approved):

PR #351, opened 2022-08-05
#347: SelectionControlGroup: remove paddingY config

### LINKED ISSUE #347: SelectionControlGroup: remove paddingY config
This config was used to add a vertical padding before the first item, but we agreed it should always be half of the spacing between the options. So we can remove the config entry and just use `config.internalSpacing \ 2` as vertical padding.

We agreed to apply the same logic also to horizontal groups (the padding top will be half the horizontal spacing between options)

marcopiii:
Closes #347

### DIFF
--- a/packages/bento-design-system/src/CheckboxGroupField/createCheckboxGroupField.tsx
+++ b/packages/bento-design-system/src/CheckboxGroupField/createCheckboxGroupField.tsx
@@ -1,5 +1,5 @@
 import { useCheckboxGroup, useCheckboxGroupItem } from "@react-aria/checkbox";
-import { useRef } from "react";
+import { ComponentProps, useRef } from "react";
 import { FieldType } from "../Field/createField";
 import { FieldProps } from "../Field/FieldProps";
 import { createInternalCheckbox } from "../Checkbox/createInternalCheckbox";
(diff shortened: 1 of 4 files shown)


packages/bento-design-system/src/CheckboxGroupField/createCheckboxGroupField.tsx
```diff
@@ -74,8 +74,13 @@ export function createCheckboxGroupField(
             assistiveTextProps={descriptionProps}
             errorMessageProps={errorMessageProps}
           >
-            <Inset spaceY={config.group.paddingY}>
-              {(props.orientation || "vertical") === "vertical" ? (
+            <Inset
+              spaceY={
+                ((config.group.internalSpacing[props.orientation ?? "vertical"] as number) /
+                  2) as ComponentProps<typeof Inset>["spaceY"]
+              }
```
marcopiii, line 81: @veej this fails at runtime if the `spacing/2` isn't a valid spacing, but allowing a custom spacing for `Inset` would make its interface less strict and I think it isn't something we want.

veej: I think you can easily replace the `Inset` with a raw `Box` passing a `style` to bypass the design tokens

veej (approved):
Looks 👍

veej:
😕 @marcopiii, looking at the diff highlighted by Chromatic, I think I make a mistake when I migrated the issue on GitHub. For horizontal groups, the vertical padding should still be `verticalInternalSpacing / 2`, instead of `horizontalInternalSpacing / 2`.
@Phyele could you confirm this is what we decided?

veej (approved):
Looks 👍

Issue #277, opened 2022-07-15
Bug: `ts-deepmerge` is merging props of JSX.Elements passed in the config 


gabro:
### Description of the bug

Given a component config containing a JSX.Element (e.g. activeVisualElement for Navigation), ts-deepmerge is deep-merging also the JSX.Elements instead of replacing them completely.

Deep-merging two JSX.Elements means their props are merged together: see https://codesandbox.io/s/react-typescript-forked-zbvhtr?file=/src/index.tsx

### Version

0.12.0

### What browsers are you seeing the problem on?

Chrome, Safari, Firefox, Edge

PR #358, opened 2022-08-17
Replace constructor approach with context + module agumentation

gabro:
This PR rethinks how we construct Bento components and I believe it carries several benefits to both Bento maintainers and Bento users.

Right now, we rely on a constructor approach where we pass around the sprinkles function and potentially other parameters, such as Chip's `customColors`.
The main motivation for this approach is that these custom parameters must be reflected in the types, so add a bunch of generics and we let these parameters drive the type inference.

While this approach works, I think it has several drawbacks, so in this PR I propose an alternative approach, based on context and module augmentation.

Let's see a few examples of usage first:

### Current

```tsx
import { createBentoComponents } from "@buildo/bento-design-system"

const { BentoProvider, Card } = createBentoComponents()

<BentoProvider ...>
  <Card ... />
</BentoProvider>
```

### New 

```tsx
import { BentoProvider, Card } from "@buildo/bento-design-system"

<BentoProvider ...>
  <Card ... />
</BentoProvider>
```

### Current

```tsx
import { createBentoComponents } from "@buildo/bento-design-system"

const { BentoProvider, Card } = createBentoComponents({
  card: { defaultRadius: 16 }
})

<BentoProvider ...>
  <Card ... />
</BentoProvider>
```

### New

```tsx
import { BentoProvider, Card } from "@buildo/bento-design-system"

<BentoProvider config={{ card: { defaultRadius: 16 }}}>
  <Card ... />
</BentoProvider>
```

or alternatively

```tsx
import { BentoProvider, Card, createBentoProvider } from "@buildo/bento-design-system"

// This is useful when you have a separate design-system module, which can use the constructor to provide a default config
const BentoProvider = createBentoProvider({
  card: { defaultRadius: 16 }
})

<BentoProvider>
  <Card ... />
</BentoProvider>
```

### Current

```tsx
import { createBentoComponents } from "@buildo/bento-design-system"
import { sprinkles } from "./sprinkles.css"

const { BentoProvider, Box } = createBentoComponents(sprinkles)

<BentoProvider ...>
  <Box background="customBackgroundToken" />
</BentoProvider>
```

### New

```tsx
import { BentoProvider, Box } from "@buildo/bento-design-system"
import { sprinkles } from "./sprinkles.css"

<BentoProvider sprinkles={sprinkles}>
  <Box background="customBackgroundToken" />
</BentoProvider>

// This is normally defined in a separate file bento.d.ts
declare module "@buildo/bento-design-system" {
  interface TypeOverrides {
    SprinklesFn: typeof sprinkles;
  }
}
```

---

The main drawback of this approach is that users have to use module augmentation, in addition to pass in an additional parameter, but this happens once or twice per project (and only for those projects which customize the tokens), and it's something that you do and forget.

On the other hand, here's the advantages I see:

For the users:
- Much easier to get started, you can just import components and use them
- The approach is aligned with other mainstream libraries, so it's also more familiar to users
- Custom sprinkles are supported ubiquitously without the need of specific support from the library. E.g. if you define a custom border radius token, you can just use it in components which accept a border radius as prop (like `Card`) 

For us:
- The codebase is much simpler, since we don't need to wrap everything into constructors to access the config
- Components can just import other components, no need to add them as a parameter to the constructor. This also means we don't need to distinguish between "internal" and "external" components.
- Docs generation (for the website) and control generation (for storybook) works much better now, since we can just point react-docgen-typescript to the library `index.ts` and be done with it
- Adding new customizable generics (similar to the one in `Chip`) does not require big refactors and "generic drilling"

---

I've tried to keep the commits small, but the diff isn't particularly helpful since most components have been renamed too (from `createX` to `X`)


packages/bento-design-system/src/BentoConfigContext.tsx
```diff
@@ -0,0 +1,25 @@
...
+}: {
+  value: PartialBentoConfig;
+  children: Children;
+}) {
+  return (
+    <BentoConfigContext.Provider value={merge(defaultConfigs, config)}>
+      {children}
+    </BentoConfigContext.Provider>
+  );
+}
```
gabro, line 25: This is where we merge default and user config, and propagate it down via context

packages/bento-design-system/src/BentoProvider.tsx
```diff
@@ -32,28 +35,40 @@ type Props = {
...
 
     return (
       <I18nProvider locale={locale}>
         <OverlayProvider style={{ height: "100%" }}>
           <DefaultMessagesContext.Provider value={{ defaultMessages }}>
-            <LinkComponentContext.Provider value={linkComponent ?? linkComponentFromContext}>
-              <ToastProvider dismissAfterMs={toastDismissAfterMs}>{children}</ToastProvider>
-            </LinkComponentContext.Provider>
+            <BentoConfigProvider value={props.config ?? config}>
+              <SprinklesContext.Provider value={props.sprinkles ?? sprinkles}>
```
gabro, line 61: we allow `BentoProvider` to override `config` and `sprinkles` via props

packages/storybook/stories/index.tsx
```diff
@@ -132,20 +131,24 @@ export const {
...
   {
-    ...defaultConfigs.actions,
-    buttonsAlignment: "spaceBetween",
+    actions: {
+      ...defaultConfigs.actions,
+      buttonsAlignment: "spaceBetween",
+    },
   },
-  { Button, Banner, InlineLoader }
+  Actions
```
gabro, line 153: `withBentoConfig` is a utility for wrapping a component in a custom `BentoConfigProvider`, so that we can create different variants of a component.

packages/bento-design-system/src/BentoConfigContext.tsx
veej: ain't we solving https://github.com/buildo/bento-design-system/issues/277 here, right?

packages/website/docs/03-Customization/03-atoms-augmentation.mdx
```diff
@@ -96,15 +96,32 @@ The `background` CSS property is part of the statusProperties, since we can spec
...
 
 ```ts title="my-project/design-system/src/index.ts"
-import { createBentoComponents } from "@buildo/bento-design-system";
+import { createBentoProvider, PartialBentoConfig } from "@buildo/bento-design-system";
 import { mySprinkles } from "./mySprinkles.css";
 
-export const { Box } = createBentoComponents(mySprinkles);
+const config: PartialBentoConfig = {}; // empty config, just for the sake of the example
+
+export const { Box } = createBentoProvider(config, mySprinkles);
```
veej, line 107: is this correct? I think it should be:
```suggestion
export const BentoProvider = createBentoProvider(config, mySprinkles);
```

veej (changes requested):
Quickly skimmed all the components and reviewed the relevant parts and looks 💯 
Just a small issue in a documentation example.

Also, is this PR braking, right?

gabro:
> Also, is this PR braking, right?

Yes, it's technically breaking because I've removed the constructors of the single components and also invoking `createBentoComponents` multiple times won't exactly work the same way as before (you would need to use the `BentoProvider` return by each call for it work).

That said, I've tried to minimize the breaking changes by keeping `createBentoComponents` around. Basically the impact of the changes should be the same as the one in the `storybook/stories/index.tsx` file in this PR (i.e. using `withBentoConfig` instead of `createBentoComponents` for custom variants and adding the module augmentation explicitly).

packages/bento-design-system/src/Chip/Config.ts
```diff
@@ -14,7 +14,7 @@ export type ChipConfig<AtomsFn extends typeof bentoSprinkles, CustomColor extend
   spacingAfterIcon: BentoSprinkles["gap"];
   spacingAfterLabel: BentoSprinkles["gap"];
   customColors: {
-    [k in CustomColor]: BoxProps<AtomsFn>["background"];
+    [k in ConfiguredTypes["ChipCustomColors"] & string]: BentoSprinkles["background"];
```
gabro, line 17: this is how we replaced the generic

packages/bento-design-system/src/SprinklesContext.ts
```diff
@@ -0,0 +1,9 @@
+import { createContext, useContext } from "react";
+import { bentoSprinkles } from "./internal";
+import { SprinklesFn } from "./util/SprinklesFn";
+
+export const SprinklesContext = createContext<SprinklesFn>(bentoSprinkles);
+
+export function useSprinkles() {
+  return useContext(SprinklesContext);
+}
```
gabro, line 9: This is how sprinkles are propagated through the library

packages/bento-design-system/src/createBentoComponents.ts
```diff
@@ -40,191 +10,100 @@ import {
...
-
-  const Tabs = createTabs(merge(defaultConfigs.folderTabs, config.tabs ?? {}));
-
-  const { Toast, ToastProvider } = createToast(merge(defaultConfigs.toast, config.toast ?? {}), {
-    Button,
-    IconButton,
-  });
-
-  const DesignSystemProvider = createBentoProvider(ToastProvider);
+function internalCreateBentoComponents(
```
gabro, line 102: Here I've tried to preserve the same interface as before. It should be unnoticeable if you just call `createBentoComponents` once.

It will be different if you call it more than once with different configurations, since it then attaches the config to `BentoProvider` and not to the components themselves, so you need to either use `withBentoConfig` or wrap the component in the respective `BentoProvider`.

Not a big deal anyway, it's a simple migration.

packages/storybook/stories/bento.d.ts
```diff
@@ -1,7 +1,10 @@
 import "@buildo/bento-design-system";
+import { sprinkles } from "./sprinkles.css";
 
 declare module "@buildo/bento-design-system" {
   interface TypeOverrides {
     LocalizedString: StrictLocalizedString;
+    SprinklesFn: typeof sprinkles;
+    ChipCustomColors: "custom";
```
gabro, line 8: example usage for the module augmentation approach

packages/storybook/stories/index.tsx
```diff
@@ -2,10 +2,9 @@ import "@buildo/bento-design-system/lib/index.css";
 import "@buildo/bento-design-system/lib/defaultTheme.css";
```
gabro, line 2: The changes in this file should be representative of the breaking changes I expect to see in the projects when upgrading

packages/website/docusaurus.config.js
```diff
@@ -62,7 +62,7 @@ const config = {
     [
       "docusaurus-plugin-react-docgen-typescript",
       {
-        src: ["../storybook/stories/index.tsx"],
+        src: ["../bento-design-system/src/index.ts"],
```
gabro, line 65: so much easier!

packages/bento-design-system/src/BentoConfigContext.tsx
gabro: I don't expect this change to have any effect on it, no.

PR #359, opened 2022-08-17
#354: Menu: allow sub-menus

### LINKED ISSUE #354: Menu: allow sub-menus
## Description

Currently, the `Menu` item doesn't allow to have sub-menus, opened on hover/click of a menu item, like:

Image

It's currently possible to achieve a similar result using some workarounds ([playroom](https://playroom.bento.buildo.io/#?code=N4Igxg9gJgpiBcJgApkEoAEBeAfB4AOgHYYYBOMALgK5knLGmkA8OjTLAsjEdex6QDOASwBeMLARAAbAIZkA5jCn8BwyjAC2grMADaqgfkNGMcgEYxp8DAxKmjzbrxMOhYiVLmLlIV2-UtHX1-N3wzWUtrDClBanNAzQBGKQAaDAgiAAUKQUEbdGw8YABfDBLU0LdgCKibWPjEgCY0jOzc-NtMXHwyiqqjAF0SgYEABzkwLR5KSRAyYQUAC0oMSggxlXswjkoFhSUyXWQKADN0vcXDnI3BdJr1g%2BkYcu68Ox3TZgAhCAAPUYOQSUACez10NU08gUwiI9RAAFokgA2MYAkDpMayKBQWEKGwo8ojbafARnXRnYmk0zAAB09MuBxgZBuY0EVOpTDYJM5GAAkhpNBgUjzScwAPS-P7c3loDnUiCnU6CKi6ADMAAZ5TtxTKdmhKqKjJkcjA8gU3r1DZ89rJhNI8XzIHD%2Bc6AMJLGAANzImQASosVta3P0jUxCGGOBYrAVARgnDw%2BJHTCJxHNvEotrzEsEDMmHDVo9EGglBWrWiaOhair1ysHeYXIjGYiA4qWtAAWCvtM2dQo9Up1uOkYbDjATWRTTQzOb7FZrDZZzmMw7HM4XfbXX1s%2B4Lp4vEqWj68%2BNSsccYFgiTASHQ2HwpGo9GY7G4oj44XIonnpjk4CUn9SDpBlN2ZVl2R-PUT35QUMBafMvklf4oNJOUx0VZVVWATVtTCXVRgNUZK17asB1DG0yDtB13ydTIbFoogPW9X0iADZZKHrUxyNMUcwxXZk1xgc41lAlltzucJHgUZ5XhrY8HB%2Bf5Rj-ACENxQQJxBOZYWomAEXMaQIDAABrJcwkyN0HRM45LSk550Fw0wwFoQQICOKQxggWENDIMzqnpWl%2BLE25HI4FCFIAQS9WRKHkDBIEM9yQC9YQIGeSgpAwfDkwlKVwqYNCwwwlVZmAJItX8XEUtgMhBBMbKOAlPU0AAbmIQ8HOIDEQAAd2EKBKCWQQED0cqmg7YYgA)), consisting in:
- passing a dummy `onPress` event to the menu item, so that its "interactive" style is used (i.e. background style on hover)
- passing a Menu as the item label, representing the sub-menu for that item
- as a trigger for the internal Menu (i.e. the sub-menu), pass a Box with negative margin and internal padding, in order to compensate the padding assigned to every Menu item

What we would like to do instead is:
- allowing to pass nested items to the Menu, e.g.
```
items={[
   { label: "Item1", onPress: () => {...} },
   { label: "Item2", subitems: [...] }
]}
```
- pre-process the list of items so that, instead of passing them directly to the `List` component underneath, every item with subitems is converted into a list item with an onPress/[onHover*] that handles the logic to open the submenu
- the submenu is rendered with an additional Popover internally to the `Menu` component (only one sub-menu open at a time)
- if we want to handle also the sub-menu open on hover, we must also adapt the `ListItem` component to accept hover handlers (onMouseEnter/onMouseLeave?). When passed, they should activate the "interactive" style (same as onPress/href).
   - pay attention to the behaviour when the mouse moves from the item to the submenu: the submenu should not be closed even if mouse is leaving the menu item.

federico-ercoles:
~~At the moment, this PR implements first-level sub-menus for the `Menu` component.~~
This PR enables the `Menu` component to have sub-menus as items, that can be opened on click.

### tests performed
✅ Sub-menus open and close correctly
✅ Sub-menus are positioned correctly in respect to the corresponding main-menu element


packages/bento-design-system/src/Menu/createMenu.tsx
```diff
@@ -119,6 +177,157 @@ export function createMenu(
...
+      if (!isSelected) {
+        childMenuState.close();
+        setSelectedMenuItem(undefined);
+      }
+    }, [isSelected, childMenuState, setSelectedMenuItem]);
+
+    const nestedMenuPortal = useMemo(
+      () =>
+        createPortal(
+          <FocusScope restoreFocus>
```
federico-ercoles, line 236: Currently focus may not be working perfectly: when opening a sub-menu, `Tab` does not immediately shift focus to the first element of the newly opened sub-menu, but finishes cycling through the elements of the currently selected menu before shifting to the opened sub-menu. I'm not sure which behaviour we'd rather have: intuitively I assume the first, but I couldn't find an easy way to implement it with `FocusScope`.

federico-ercoles:
@gabro Ready for rebase

veej (commented):
The behaviour looks quite ok, even if sometimes the submenu takes a bit longer to open (not sure if it's my mac's fault, or something in the implementation).

I left a few comments inline, and I would also like to hear @gabro's opinion, since the implementation is quite complex.

packages/bento-design-system/src/List/ListItem.tsx
```diff
@@ -54,29 +55,31 @@ type RightItem = {
...
-      }
-  );
+type CommonProps = {
+  disabled?: boolean;
+  size: ListSize;
+  isFocused?: boolean;
+  isSelected?: boolean;
+  ignoreTabIndex?: boolean;
+  key?: string | number;
+  ref?: RefObject<HTMLElement>;
```
gabro, line 65: I don't think this is necessary, `forwardRef` automatically adds `ref` to props.

packages/bento-design-system/src/Popover/Popover.tsx
```diff
@@ -29,50 +29,48 @@ type Props = {
...
-      isOpen: true,
-      isDismissable: true,
-      onClose,
-    },
-    overlayRef
-  );
+export const Popover = forwardRef<HTMLElement, Props>(
+  ({ children, onClose, triggerRef, placement = "bottom start", offset = 4 }, ref) => {
+    const overlayRef = useRef<HTMLElement>(null);
+    useImperativeHandle(ref, () => overlayRef.current as HTMLElement);
```
gabro, line 35: why is the cast necessary? `overlayRef.current` should be `HTMLElement`

gabro (changes requested):
Apart from @veej's comments and some other comments I've left inline, I think there's one major thing to tackle: we're essentially hacking around our own `ListItem` component. I think we shouldn't.

Since `Menu` is "outgrowing" `ListItem` in terms of complexity, I think we should use a dedicate component for the menu items (we can re-use style recipes and config from `ListItem` so to keep visual consistency).

This should simplify the implementation a bit.

Another minor thing before merging is a name cleanup, there are least three names used for sub-menus: sub, child and nested. Let's pick one and stick to it.

packages/bento-design-system/src/Popover/Popover.tsx
federico-ercoles: Because of the definition at line 34, `overlayRef.current` is actually `HTMLElement | null`.

federico-ercoles:
The latest commits I pushed address most of your feedbacks, except for the refactor to move away from using `ListItem` and those regarding refs (which would be easier to address in the refactor), due to the time constraint of me being out of office next week.

veej:
@gabro @federico-ercoles I tried to use `InternalList` and a custom `NestedMenuListItem`, and it looks way easier this way. Have a look at the last commit!

federico-ercoles:
> @gabro @federico-ercoles I tried to use `InternalList` and a custom `NestedMenuListItem`, and it looks way easier this way. Have a look at the last commit!

After a quick check, visually everything looks ok, but it seems nested menus cannot be toggled open and closed any more (i.e. I can open a nested menu by clicking the list item, but I cannot close it by clicking again the list item, I can only close it by opening another nested menu at the same level or above).

veej (approved):
💣

gabro (approved):

PR #361, opened 2022-08-20
Migrate storybook to new 0.13 API and cleanup formatMessage

gabro:
~Migrate the storybook to the recently introduced API (no more `createBentoComponents`).~

As a bonus, I've also removed all uses of `formatMessage` since it became useless after #165.

### DIFF
--- a/packages/storybook/stories/index.tsx
+++ b/packages/storybook/stories/index.tsx
@@ -58,7 +58,6 @@ export {
   CustomModal,
   DateField,
   DecorativeDivider,
-  DesignSystemProvider,
   Disclosure,
   DisclosureGroup,
   Display,
--- a/packages/storybook/stories/index.tsx
+++ b/packages/storybook/stories/index.tsx
@@ -72,6 +72,7 @@ export {
   InlineLoader,
   Inset,
   Label,
+  Link,
   List,
   ListItem,
   Menu,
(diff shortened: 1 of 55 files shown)


veej (approved):
Looks ⭐

veej:
@gabro it looks like something is broken in storybook/chromatic :(

veej:
@gabro can we rebase this and check that everything is working correctly, so that we can formalise the new version?

veej:
still failing 😞

gabro:
It seems that I've accidentally broken the automatic `on...` insertion of props.

veej:
> It seems that I've accidentally broken the automatic `on...` insertion of props.

Yep, I tried to understand why, but I didn't find any clue.

packages/storybook/stories/index.tsx
```diff
@@ -35,6 +35,21 @@ const FeedbackBackground = (
...
+//
+// The workaround would be to wrap each component in another component that just
+// re-exports it (see https://github.com/storybookjs/storybook/issues/13502#issuecomment-752897475),
+// but that would be way too inconvenient.
+//
+// Using createBentoComponents creates the components directly in Storybook, so
+// their props are picked up correctly.
+//
+// We tried several solutions from the linked issues, but it seems our monorepo
+// setup make this particularly tricky
```
gabro, line 52: @veej I couldn't find a workaround for this, so I've at least documented the whole thing

PR #365, opened 2022-08-29
Remove ts-toolbelt

gabro:
Having ts-toolbelt in the public interface seems to be causing problems with typings downstream


veej (approved):
looks 💯

PR #366, opened 2022-08-29
Remove duplicate FieldProps eport

gabro:
We accidentally exported also the props of `Field` as `FieldProps`, which would then shadow the original `FieldProps`.

I've opted for removing the export since it doesn't feel very useful and it can be always retrieved as `ComponentProps<typeof Field>` if needed


federico-ercoles (approved):
🚀

PR #383, opened 2022-09-30
#381: Certain navigation components don't trigger their action

### LINKED ISSUE #381: Certain navigation components don't trigger their action
Some components (`Link`, `ButtonLink`, `Navigation` and `Breadcrumb`, that I've seen so far) don't perform their action when clicked.

federico-ercoles:
Closes #381

Components acting as `LinkComponent` were not working correctly, because they were not handled properly in `Box`.
That was happening because of this check:
```typescript
    const el = typeof element === "string" ? element : "div";
```
which, before the refactor, was only used to determine which `resetStyle` was to be applied to the element. After the refactor, this check also determined what element was created. When `as={LinkComponent}` was passed to `Box`, this check would return `div` and therefore it wouldn't behave as a link.

## Test Plan
✅ Components depending on `LinkComponent` work correctly


gabro (approved):

gabro:
Nice catch!

PR #396, opened 2022-10-10
Use sprinkles from context instead of bentoSprinkles

### LINKED ISSUE #395: Bug: Custom sprinkles are overridden by the default ones in some components
### Description of the bug

Some components are using the default `bentoSprinkles` instead of getting them via `useSprinkles`. For example
https://github.com/buildo/bento-design-system/blob/5acaf1cd8114d23435b0933ad16d456c8fdcba55/packages/bento-design-system/src/SelectField/components.tsx#L162-L168

This causes runtime errors if you try to use a custom sprinkle.

### How to reproduce
Define a custom radius
```
// theme.css.ts
export const customVars = createGlobalTheme(":root", {
  borderRadius: {
    0: "0px",
  },
});

// sprinkles.css.ts
const defineBentoSprinkles = createDefineBentoSprinklesFn();
export const { sprinkles: customSprinkles } = defineBentoSprinkles({
  ...unconditionalProperties,
  borderRadius: {
    ...unconditionalProperties.borderRadius,
    ...customVars.borderRadius,
  },
});
export type CustomSprinkles = Parameters<typeof customSprinkles>[0];
```

Use the custom radius in the dropdown config
```
createBentoProvider(
  {
    dropdown: {
      radius: 0
    },
  },
  customSprinkles
);
```

Run the showroom, and open the dropdown to get 
```
SprinklesError: "borderRadius" has no value 0. Possible values are "4", "8", "16", "circled", "circledX"
```

### Playroom reproduction

_No response_

### Version

0.13.6

Chrome

gabro:
Fixes #395

### DIFF
--- a/packages/bento-design-system/src/AreaLoader/AreaLoader.tsx
+++ b/packages/bento-design-system/src/AreaLoader/AreaLoader.tsx
@@ -1,5 +1,4 @@
-import clsx from "clsx";
-import { bentoSprinkles, Box, Inline, Stack, Body, LocalizedString } from "..";
+import { Box, Inline, Stack, Body, LocalizedString } from "..";
 import { container, dot, text } from "./AreaLoader.css";
 import { BentoSprinkles } from "../internal";
 import { AreaLoaderConfig } from "./Config";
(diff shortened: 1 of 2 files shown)


packages/bento-design-system/src/AreaLoader/AreaLoader.tsx
```diff
@@ -54,28 +53,24 @@ export type { Props as AreaLoaderProps };
...
   return (
     <Box
-      className={clsx(
-        container,
-        bentoSprinkles({
-          background: config.scrimColor === "dark" ? "backgroundDarkScrim" : "backgroundLightScrim",
-        })
-      )}
+      background={config.scrimColor === "dark" ? "backgroundDarkScrim" : "backgroundLightScrim"}
+      className={container}
```
gabro, line 60: it didn't event make sense to use `sprinkles` directly here, since we can just use `Box` (which internally already uses `sprinkles` from the context)

gabro: basically it makes sense to use `sprinkles` directly only if we can't use `Box` directly, and the only case we have right now is the custom styling of react-select components.

marcopiii (approved):

PR #399, opened 2022-10-10
Add 0 to radius allowed values

gabro:
See https://github.com/buildo/bento-design-system/discussions/394


packages/bento-design-system/src/util/atoms.ts
```diff
@@ -2,6 +2,7 @@ import { vars } from "../vars.css";
 
 const radius = {
   ...vars.radius,
+  0: 0,
```
gabro, line 5: added here and not in the theme, otherwise it would be a breaking change.

(@veej heads up, I'm planning to remove a bunch of variables from the theme since I don't think they belong there, radius is a good example)

PR #405, opened 2022-10-13
Make folder Tabs active background/foreground colors configurable

gabro:


gabro:
Discussed with @Phyele, we don't actually want to make this configurable so I'll close it

PR #408, opened 2022-10-13
Make it easier to customize the default Navigation active element

gabro:


veej (approved):
Great 💣 
Does this also limit the impact of https://github.com/buildo/bento-design-system/issues/277?
Or, what's the behavior when we try to merge the current default object with a JSX.Element?

marcopiii (approved):
🍍

PR #410, opened 2022-10-14
Remove unused iconSize from InputConfig

### LINKED ISSUE #370: InputConfig's iconSize seems unused
I've tried removing it and nothing happened :D 

It was probably used at some point, but we have more specific icon sizes for inputs like `passwordIconSize` and `searchIconSize`

gabro:
Closes #370


veej (approved):
💯

PR #411, opened 2022-10-14
Deprecate IllustrationProps "style" in favor of "kind"

### LINKED ISSUE #371: IllustrationProps "style" attribute raises a jsx-eslint warning
The name "style" of the attribute in https://github.com/buildo/bento-design-system/blob/edfc0ebe181c280afb65a270dc3c0ad82a897780/packages/bento-design-system/src/Illustrations/IllustrationProps.ts#L5 makes jsx-eslint to give the following warning

 https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/style-prop-object.md

gabro:
Closes #371


packages/bento-design-system/src/Illustrations/IllustrationProps.ts
```diff
@@ -1,11 +1,22 @@
+// NOTE(gabro): we deprecated the `style` prop due to false positive warnings in
+// popular ESLint configs, which expect `style` to be an object.
+// We're still keeping the old props around for backwards compatibility.
 export type IllustrationProps = {
   size: 24 | 32 | 40 | 80 | 160 | { custom: number };
 } & (
   | {
-      style: "color";
+      kind: "color";
```
veej, line 8: isn't it a breaking change in this way? I mean, the `kind` prop looks mandatory, so one has to add the kind everywhere, so what's the point in keeping also the style?

gabro: Yeah, it was broken, fixed it in my last commit!

veej (approved):
💯

PR #420, opened 2022-11-03
Avoid including react-dom in the bundle

gabro:
While doing some experiments on the build I've noticed we were accidentally including react-dom in the bundle.

This is because tsup automatically excludes `dependencies` and `peerDependencies`, but `react-dom` was only added as `devDependency` (which is fine, for development) and not as `peerDependency`!

This PR basically cuts the final js bundle by 40% 🤦


veej (approved):
💣

PR #432, opened 2022-11-08
Uniform the `color` prop of typography components

### LINKED ISSUE #421: Add support for colors in Display and Headline
Currently, the components `Display` and `Headline` don't support the property `color` (they always use the primary color).

The other typography components support the following colors

|                  | Body         | Display      | Headline     | Label | Title        |
| ---------------- | ------------ | ------------ | ------------ | ----- | ------------ |
| Primary          | ✅ (as `default`) | ✅ (implicitly) | ✅ (implicitly) | ✅    | ✅ (as `default`) |
| Secondary        | ✅           |              |              | ✅    | ✅           |
| PrimaryInverse   | ✅           |              |              | ✅    | ✅           |
| SecondaryInverse |              |              |              | ✅    | ✅           |
| Interactive      |              |              |              |       |              |
| Informative      | ✅           |              |              | ✅    | ✅           |
| Positive         | ✅           |              |              | ✅    | ✅           |
| Warning          | ✅           |              |              | ✅    | ✅           |
| Negative         |              |              |              | ✅    | ✅           |
| Disabled         | ✅           |              |              | ✅    |              |

marcopiii:
Closes #421 

This PR uniforms the `color` prop definition of the typography components.
The supported colors are now the following

| | Body | Display | Headline | Label | Title |
| - | - | - | - | - | - |
| Primary | ✅ | ✅ | ✅ | ✅ | ✅ |
| Secondary | ✅ | ✅ | ✅ | ✅ | ✅ |
| PrimaryInverse | ✅ | ✅ | ✅ | ✅ | ✅ |
| SecondaryInverse | ✅ | ✅ | ✅ | ✅ | ✅ |
| Interactive | ✅ |  |  | ✅ |  |
| Informative | ✅ |  |  | ✅ | ✅ |
| Positive | ✅ |  |  | ✅ | ✅ |
| Warning | ✅ |  |  | ✅ | ✅ |
| Negative | ✅ |  |  | ✅ | ✅ |
| Disabled | ✅ |  |  | ✅ |  |

All the components also support the `inherit` value.
The default value is' primary' for all the components except `Label` (to not alter its behaviour).


gabro (changes requested):
Minor concern about the breaking changes we're introducing in Body.

Also, every change to the recipes in technically breaking since those are exposed as public API, but I'm less concerned about those, even though it would be nice to avoid it.
Maybe we can wrap those recipes and make color optional?

Let's discuss it before merging

marcopiii:
I understand the concern about the breaking change in the recipes, but I'm not that sure that preserving retro compatibly, in this case, is worth introducing an inconsistency between the recipes, where `color` is mandatory somewhere and optional elsewhere.

gabro (approved):
Discussed offline, let's go with the breaking change since the impact is fairly small and the migration is easy

PR #434, opened 2022-11-09
Improve underline configuration

### LINKED ISSUE #423: Improve line configuration in tabs and navigation
In the Tabs component, the `lineHeight` accepts only the values of `borderBottomWidth`, while it should probably accept a `SizeConfig<number>` like the `activeVisualElement` of the Navigation component.

In the Tabs component, the `lineColor` accepts the values of "borderColor", while in the Navigation component the `lineColor` accepts the values of "background". We should probably unify them.

marcopiii:
Close #423

This PR is breaking since the Navigation's underline color is narrowed down from 
```
brandColor | backgroundColor | interactiveBackgroundColor | outlineColor | dataVisualizationColor 
```
to
```
brandColor | foregroundColor
```
and the Tab's config changes from
````
lineHeight: BentoSprinkles["borderBottomWidth"];
````
to 
```
lineHeight: SizeConfig<number>;
```

### DIFF
--- a/packages/bento-design-system/src/util/atoms.ts
+++ b/packages/bento-design-system/src/util/atoms.ts
@@ -89,6 +89,7 @@ const color = {
   ...vars.interactiveForegroundColor,
   ...vars.outlineColor,
   ...vars.dataVisualizationColor,
+  transparent: "transparent",
 };
 
 const background = {
@@ -99,8 +100,15 @@ const background = {
   ...vars.dataVisualizationColor,
 };
 
+const decoration = {
+  ...vars.brandColor,
+  ...vars.foregroundColor,
+  transparent: "transparent",
+};
+
 export const statusProperties = {
   color,
+  decoration,
   background: { ...background, currentColor: "currentColor" },
   cursor: {
     pointer: "pointer",
@@ -120,7 +128,7 @@ export const statusProperties = {
   stroke: color,
   textDecoration: ["none", "underline"],
   fill: { ...color, ...background, inherit: "inherit", currentColor: "currentColor" },
-  borderColor: { ...color, transparent: "transparent" },
+  borderColor: color,
   borderStyle: {
     solid: "solid",
     dashed: "dashed",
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/bento-design-system/src/util/atoms.ts [hunk 1]
```diff
@@ -89,6 +89,7 @@ const color = {
   ...vars.interactiveForegroundColor,
   ...vars.outlineColor,
   ...vars.dataVisualizationColor,
+  transparent: "transparent",
```
marcopiii, line 92: we had to add `transparent` to `color` to be able to pass such value to the navigation's `Box`

veej: @gabro in order to use the a foreground color as a background, we're using the `background="currentColor"` trick. This means the decorator color is passed to the `color` Box prop, but decorator colors support the additional "transparent" value that is not supported by `color`. Thus, we added it to the available colors too, meaning that now it's theoretically possible to use "transparent" as a foreground color. Anything against it?

packages/bento-design-system/src/util/defaultConfigs.tsx
```diff
@@ -483,7 +483,10 @@ export const underlineTabs: TabsConfig = {
     medium: 16,
     large: 32,
   },
-  lineHeight: 2,
+  lineHeight: {
+    medium: 2,
+    large: 4,
+  },
```
veej, line 489: is this a leftover, or is it deliberate?

veej (approved):
Just a doubt inline, and I'd wait @gabro's opinion on that "transparent" color

packages/bento-design-system/src/util/defaultConfigs.tsx
marcopiii: deliberate: we want to be able to define different widths based on the size of the component.

gabro: It's ok to allow specifying a different height, but do we really want to change the value too? The more conservative option would be 
```suggestion
  lineHeight: {
    medium: 2,
    large: 2,
  },
```

gabro (commented):
One question

gabro (approved):

PR #442, opened 2022-11-15
Basic charts

gabro:
`BarChart`, `LineChart` and `DonutChart`

### DIFF
--- a/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
+++ b/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
@@ -13,6 +13,7 @@ import { bodyRecipe } from "../../Typography/Body/Body.css";
 import { allColors } from "../../util/atoms";
 import { Children } from "../../util/Children";
 import { ChartDataColor } from "../Config";
+import { legendContent } from "../Legend/Legend";
 
 type Props<D extends string, C extends string> = {
   data: Record<D | C, unknown>[];
@@ -73,7 +74,7 @@ export function BarChart<D extends string, C extends string>({
         {showXAxis && <XAxis dataKey={dataKey} />}
         {showYAxis && <YAxis />}
         {showTooltip && <Tooltip />}
-        {showLegend && <Legend />}
+        {showLegend && <Legend content={legendContent} />}
         {categories.map((category, i) => (
           <Bar
             key={category}
--- a/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
+++ b/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
@@ -12,8 +12,10 @@ import { useBentoConfig } from "../../BentoConfigContext";
 import { bodyRecipe } from "../../Typography/Body/Body.css";
 import { allColors } from "../../util/atoms";
 import { Children } from "../../util/Children";
+import { vars } from "../../vars.css";
 import { ChartDataColor } from "../Config";
 import { legendContent } from "../Legend/Legend";
+import { tooltipContent } from "../Tooltip/Tooltip";
 
 type Props<D extends string, C extends string> = {
   data: Record<D | C, unknown>[];
@@ -73,7 +75,12 @@ export function BarChart<D extends string, C extends string>({
       <RechartBarChart data={data}>
         {showXAxis && <XAxis dataKey={dataKey} />}
         {showYAxis && <YAxis />}
-        {showTooltip && <Tooltip />}
+        {showTooltip && (
+          <Tooltip
+            content={tooltipContent}
+            cursor={{ fill: vars.backgroundColor.backgroundSecondary }}
+          />
+        )}
         {showLegend && <Legend content={legendContent} />}
         {categories.map((category, i) => (
           <Bar
--- a/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
+++ b/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
@@ -34,7 +34,7 @@ export function BarChart<D extends string, C extends string>({
   hideYAxis = false,
   hideLegend = false,
   disableAnimation = false,
-  hideTooltip = true,
+  hideTooltip = false,
   width = "100%",
   height,
   minWidth,
@@ -63,21 +63,21 @@ export function BarChart<D extends string, C extends string>({
       debounce={debounce}
     >
       <RechartBarChart data={data}>
-        {hideXAxis && <XAxis dataKey={dataKey} />}
-        {hideYAxis && <YAxis />}
-        {hideTooltip && (
+        {!hideXAxis && <XAxis dataKey={dataKey} />}
+        {!hideYAxis && <YAxis />}
+        {!hideTooltip && (
           <Tooltip
             content={tooltipContent}
             cursor={{ fill: vars.backgroundColor.backgroundSecondary }}
           />
         )}
-        {hideLegend && <Legend content={legendContent} />}
+        {!hideLegend && <Legend content={legendContent} />}
         {categories.map((category, i) => (
           <Bar
             key={category}
             dataKey={category}
             fill={colors[i % colors.length]}
-            isAnimationActive={disableAnimation}
+            isAnimationActive={!disableAnimation}
             stackId={stacked ? "stack" : undefined}
           />
         ))}
--- a/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
+++ b/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
@@ -19,8 +19,8 @@ type Props<D extends string, C extends string> = ChartProps & {
   data: Record<D | C, unknown>[];
   categories: C[];
   dataKey: D;
-  showXAxis?: boolean;
-  showYAxis?: boolean;
+  hideXAxis?: boolean;
+  hideYAxis?: boolean;
   stacked?: boolean;
 };
 
@@ -30,11 +30,11 @@ export function BarChart<D extends string, C extends string>({
   data,
   dataKey,
   categories,
-  showXAxis = true,
-  showYAxis = true,
-  showLegend = true,
-  showAnimation = true,
-  showTooltip = true,
+  hideXAxis = false,
+  hideYAxis = false,
+  hideLegend = false,
+  disableAnimation = false,
+  hideTooltip = true,
   width = "100%",
   height,
   minWidth,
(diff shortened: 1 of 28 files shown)


packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { bodyRecipe } from "../../Typography/Body/Body.css";
+import { allColors } from "../../util/atoms";
+import { vars } from "../../vars.css";
+import { ChartProps } from "../ChartProps";
+import { legendContent } from "../Legend/Legend";
+import { tooltipContent } from "../Tooltip/Tooltip";
+
+type Props<D extends string, C extends string> = ChartProps & {
+  data: Record<D | C, unknown>[];
+  categories: C[];
```
veej, line 20: can't we just infer the categories from the data keys? Not sure if this would break the inference though

packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export function BarChart<D extends string, C extends string>({
+  data,
+  dataKey,
+  categories,
+  showXAxis = true,
+  showYAxis = true,
+  showLegend = true,
+  showAnimation = true,
+  showTooltip = true,
```
veej, line 37: just a little obsession of mine: since we're defaulting these to `true`, wouldn't it be better to invert them and call them `hideXXX` instead, so to default them to `false`? In this way, the default is = to not passing any prop, and to disable them you just need to pass `hideXXX`, instead of `showXXX={false}`

packages/bento-design-system/src/Charts/Config.ts
```diff
@@ -0,0 +1,7 @@
+import { allColors } from "../util/atoms";
+
+export type ChartDataColor = keyof typeof allColors;
```
veej, line 3: shouldn't we use only the dataViz colors here?

packages/bento-design-system/src/Charts/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,83 @@
...
+} from "recharts";
+import { useBentoConfig } from "../../BentoConfigContext";
+import { bodyRecipe } from "../../Typography/Body/Body.css";
+import { allColors } from "../../util/atoms";
+import { ChartProps } from "../ChartProps";
+import { legendContent } from "../Legend/Legend";
+import { tooltipContent } from "../Tooltip/Tooltip";
+
+type Props<D extends string, C extends string> = ChartProps & {
+  data: Record<D, unknown>[];
```
veej, line 17: shouldn't this include also `C` as keys?
```suggestion
  data: Record<C | D, unknown>[];
```

packages/bento-design-system/src/Charts/Legend/Legend.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Body } from "../../Typography/Body/Body";
+
+export const legendContent: Required<RechartsLegendProps>["content"] = ({ payload = [] }) => {
+  return (
+    <Box paddingTop={8} width="full">
+      <Inline space={16} alignY="center" align="center">
+        {payload.map((entry) =>
+          match(entry.type ?? "square")
+            .with("plainline", "line", () => <LegendItemLine key={entry.value} {...entry} />)
+            .otherwise(() => <LegendItemArea key={entry.value} {...entry} />)
```
veej, line 16: are there many other possible values? Can't we use an exhaustive match instead?

packages/bento-design-system/src/Charts/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <RechartLineChart data={data}>
+        {categories.map((category, i) => (
+          <Line
+            type={lineType}
+            key={category}
+            dataKey={category}
+            isAnimationActive={showAnimation}
+            stroke={colors[i % colors.length]}
+            strokeWidth={2}
+            dot={false}
```
veej, line 81: does it make sense to make this configurable?

packages/bento-design-system/src/Charts/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,43 @@
...
+      borderStyle="solid"
+      borderColor="outlineContainer"
+      borderRadius={4}
+      tabIndex={-1}
+    >
+      <Stack space={8}>
+        <Body size="medium">{label}</Body>
+        <Stack space={4}>
+          {payload.map(({ value, name, color }) => (
+            <Columns space={4} alignY="center">
```
veej, line 32: missing key?

packages/storybook/stories/Components/Charts/DonutChart.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const { defaultExport, createStory } = createComponentStories({
+  component: DonutChart,
+  args: {
+    height: 300,
+    dataKey: "name",
+    category: "value",
+    showLegend: true,
+    showTooltip: true,
+    showAnimation: true,
```
veej, line 12: ain't these enabled by default?

packages/website/docs/04-Components/LineChart.mdx
```diff
@@ -0,0 +1,7 @@
+import ComponentDoc from "./_ComponentDoc.mdx";
+
+<ComponentDoc
+  name="LineChart"
+  examplePath="Components/LineChart"
+  storybookStoryId="components-charts-linechart"
+/>
```
veej, line 7: no doc for the DonutChart?

packages/website/src/snippets/Components/LineChart.tsx
```diff
@@ -0,0 +1,56 @@
...
+        {
+          name: "Page G",
+          uv: 3490,
+          pv: 4300,
+          amt: 2100,
+        },
+      ]}
+    />
+  );
+}
```
veej, line 56: no snippet for the DonutChart?

veej (changes requested):
Looks great!!! 📈 📉 
Left a few comments/doubts/questions inline

packages/bento-design-system/src/Charts/Config.ts
gabro: I discussed this with Agnese, and she mentioned these can be any color. We'll probably iterate a bit more on this, allowing also special palettes for charts, but this is a good starting point I think.

packages/storybook/stories/Components/Charts/BarChart.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const { defaultExport, createStory } = createComponentStories({
+  component: BarChart,
+  args: {
+    height: 300,
+    categories: ["uv", "pv", "amt"],
+    data: [
+      {
+        name: "Page A",
+        uz: 4000,
```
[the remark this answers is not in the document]
gabro: Nice catch, thanks

packages/storybook/stories/Components/Charts/DonutChart.stories.tsx
gabro: yes they are, but if we don't include them they don't show up in controls. This is due to the fact that I'm not going through `createBentoComponents` for Charts, so react-docgen does not work (see the comment in `stories/index.tsx` for more context)

packages/website/docs/04-Components/LineChart.mdx
gabro: woops, added!

packages/website/src/snippets/Components/LineChart.tsx
gabro: added!

packages/bento-design-system/src/Charts/LineChart/LineChart.tsx
gabro: I wanted to start simple and be very prescriptive with the charts. We can always add more config in the future

packages/bento-design-system/src/Charts/Legend/Legend.tsx
gabro: there are many others, but not too many, but I can make this exhaustive.

packages/bento-design-system/src/Charts/DonutChart/DonutChart.tsx
gabro: yes, it should

packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
gabro: I've tried making this smarter, but it became quite complex for little benefit

packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
gabro: It's a good point, I've basically used the API as Tremor, but I agree it should be better to change the order

gabro:
Thanks for the review @veej! I've pushed a few more changes

packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
```diff
@@ -30,11 +30,11 @@ export function BarChart<D extends string, C extends string>({
...
-  showXAxis = true,
-  showYAxis = true,
-  showLegend = true,
-  showAnimation = true,
-  showTooltip = true,
+  hideXAxis = false,
+  hideYAxis = false,
+  hideLegend = false,
+  disableAnimation = false,
+  hideTooltip = true,
```
veej, line 37: shouldn't this be defaulted to `false`?

packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
```diff
@@ -63,21 +63,21 @@ export function BarChart<D extends string, C extends string>({
       debounce={debounce}
     >
       <RechartBarChart data={data}>
-        {showXAxis && <XAxis dataKey={dataKey} />}
-        {showYAxis && <YAxis />}
-        {showTooltip && (
+        {hideXAxis && <XAxis dataKey={dataKey} />}
```
veej, line 66: shouldn't you invert the condition here?
!hideXAxis && <...

packages/bento-design-system/src/Charts/DonutChart/DonutChart.tsx
```diff
@@ -25,9 +25,9 @@ export function DonutChart<D extends string, C extends string>({
   data,
   dataKey,
   category,
-  showLegend = true,
-  showAnimation = true,
-  showTooltip = true,
+  hideLegend = true,
+  disableAnimation = true,
+  hideTooltip = true,
```
veej, line 30: shouldn't you invert the defaults here?

veej:
@gabro I also left a few comments in the chromatic stories. Seems like the animations detection isn't working correctly, so the screenshots are taken too early, when the animation is still not completed.

packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
gabro: damn, I was multitasking too much :D

gabro:
Fix the last comments, I will look into Chromatic later today

packages/storybook/stories/Components/Charts/DonutChart.stories.tsx
veej, line 12:
```suggestion
    hideLegend: false,
    hideTooltip: false,
    hideAnimation: false,
```

veej (changes requested):
last little effort

gabro:
@veej fixed, and fixed also Chromatic (maybe 🤞)

veej (approved):
👏 🚀 🥑

PR #447, opened 2022-11-18
Add several customizations for tableColumn types + allow customzing Feedback size of empty tables

gabro:
Two main improvements:

- add more options for customizing predefined table columns. Some examples:
  - `tableColumn.text` allows customizing size, weight and color
  - `tableColumn.icon` allows customizing size and color
  - and so on

- allow customizing the size of the "no results" feedback on Table


packages/bento-design-system/src/Table/cells.tsx
```diff
@@ -62,7 +78,9 @@ export function TextWithIconCell({
     text: LocalizedString;
     tooltipContent?: Children;
   }
->) {
+> & {
+  options: Partial<Pick<BodyProps, "size" | "weight" | "color">>;
+}) {
   const icon_ = icon && icon({ size: 12 });
```
veej, line 84: doesn't it make sense to configure also the icon size?

veej (approved):
Just a small doubt, but LGTM 👍 💅

packages/bento-design-system/src/Table/cells.tsx
gabro: yes it does, pushed a change

PR #454, opened 2022-11-24
Add onChangeEnd to SliderField

gabro:
This allows detecting whether the user is actively interacting with the slider or not (it works consistently across mouse and touch).

It covers the case discussed in #417

### DIFF
--- a/packages/bento-design-system/src/Slider/Slider.tsx
+++ b/packages/bento-design-system/src/Slider/Slider.tsx
@@ -4,7 +4,7 @@ import { mergeProps } from "@react-aria/utils";
 import { VisuallyHidden } from "@react-aria/visually-hidden";
 import { SliderState } from "@react-stately/slider";
 import { FocusableElement } from "@react-types/shared";
-import { DOMAttributes, OutputHTMLAttributes, RefObject, useEffect, useRef } from "react";
+import { DOMAttributes, OutputHTMLAttributes, RefObject, useRef } from "react";
 import { Box, Column, Columns, Stack } from "..";
 import { useBentoConfig } from "../BentoConfigContext";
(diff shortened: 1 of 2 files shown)


packages/bento-design-system/src/Slider/Slider.tsx
```diff
@@ -108,18 +111,24 @@ type ThumbProps = {
...
   const {
     thumbProps: { style, ...thumbProps },
     inputProps,
+    isDragging,
   } = useSliderThumb({ index, trackRef, inputRef, isDisabled: props.disabled }, state);
   const { focusProps, isFocusVisible } = useFocusRing({ autoFocus: props.autoFocus });
 
+  useEffect(() => {
+    onDragStatusChange?.(isDragging);
+  }, [isDragging, onDragStatusChange]);
```
veej, line 130: I see from the [doc](https://react-spectrum.adobe.com/react-aria/useSlider.html) that the `useSlider` hook accepts  an `onChangeEnd` property described as follows:

> Fired when the slider stops moving, due to being let go.

Is this something that can be used to achieve a similar behaviour? Have you tried it?

veej (commented):
The solution looks 👍 
I just have a doubt on a possible alternative solution.

packages/bento-design-system/src/Slider/Slider.tsx
gabro: I have seen it but I also want to notify when the user starts dragging, so I think the current approach is more flexible

gabro: Actually, I've changed my mind. I think it's cleaner to use `onChangeEnd`, which should be enough for the use case it originated this PR

PR #456, opened 2022-11-25
Make FieldProp name and onBlur optional

gabro:
`name` and `onBlur` are historically mandatory since they come from a project were we used fields exclusively with [`formo`](https://github.com/buildo/formo) so we got them for free.

However, they're a bit annoying to use when using field components standalone. Making them optional is non-breaking so let's just do it!


packages/website/src/snippets/Components/FileUploaderField.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
```
veej, line 1: leftover?

gabro: woops, yes

PR #462, opened 2022-12-01
Remove IsMulti type parameter from SelectField

### LINKED ISSUE #376: Multioptions SelectField can't accept string[] as value
### Description of the bug

Hi @gabro and @veej ,
I tried to implement a multioptions selectField and if I set 
     value={["A", "B", "C"]}
is everything ok, otherwise if I set
     const selectedLetters : string[] = ["A", "B", "C"];
     value={selectedLetters}
there's an error saying that: 
  _Type 'string[]' is not assignable to type 'string[][]'.
  Type 'string' is not assignable to type 'string[]'.ts(2322)_

Please, can you help me?

Here's the code:

```ts
const optionsLetters = [
    {
        value: "A",
        label: "A",
        kind: "single-line",
    },
    {
        value: "B",
        label: "B",
        kind: "single-line",
    },
    {
        value: "C",
        label: "C",
        kind: "single-line",
    },
    {
        value: "D",
        label: "D",
        kind: "single-line",
    },
];

const selectedLetters = ["A", "B", "C"];

<SelectField
    name={""}
    value={selectedLetters}
    onChange={(v: string[]) => {
        console.log("selected", v);
    }}
    onBlur={() => {}}
    label="Letters"
    placeholder={""}
    options={optionsLetters}
    isMulti={true}
    multiValueMessage={(numberOfSelectedOptions: number) =>
        `${numberOfSelectedOptions} letters selected`
    }
    showMultiSelectBulkActions={true}
/>
```
        

### Playroom reproduction

_No response_

### Version

0.12.2

_No response_

gabro:
Fixes #376 

We were using `IsMulti` to mimic what react-select does internally, but it turns out that we don't need it. Using a more classic tagged union approach, we can remove a bunch of casts internally, plus we fix an issue that required users to explicitly pass the type parameters to `SelectField`.

This is *technically* breaking if users were annotating types explicitly, but since the type annotations were a workaround for a bug, and this is a fix, I won't mark it as such.

### DIFF
--- a/packages/bento-design-system/src/SelectField/SelectField.tsx
+++ b/packages/bento-design-system/src/SelectField/SelectField.tsx
@@ -22,24 +22,26 @@ export type SelectOption<A> = Omit<
   value: A;
 };
 
-type Props<A, IsMulti extends boolean> = (IsMulti extends false
-  ? FieldProps<A | undefined>
-  : FieldProps<A[]>) & {
+type MultiProps<A> = {
+  isMulti: true;
+  multiValueMessage?: (numberOfSelectedOptions: number) => LocalizedString;
+  showMultiSelectBulkActions?: boolean;
+  selectAllButtonLabel?: LocalizedString;
+  clearAllButtonLabel?: LocalizedString;
+} & FieldProps<A[]>;
+
+type SingleProps<A> = {
+  isMulti?: false;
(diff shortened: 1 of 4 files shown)


veej (approved):
💯

PR #463, opened 2022-12-02
Make placeholder optional on fields

gabro:
We were a bit too strict on requiring a placeholder on every field.
Placeholders are quite controversial in general (see https://www.smashingmagazine.com/2018/06/placeholder-attribute/) and it makes sense for us not to be prescriptive about it.

### DIFF
--- a/packages/bento-design-system/src/NumberField/NumberField.tsx
+++ b/packages/bento-design-system/src/NumberField/NumberField.tsx
@@ -10,7 +10,7 @@ import { Field } from "../Field/Field";
 import { NumberInput } from "../NumberInput/NumberInput";
 
 type Props = FieldProps<number | undefined, number> & {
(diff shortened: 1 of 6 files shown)


veej:
Based on the points the article you posted mention, I'd say we should be prescriptive on using or not using placeholders in general for a project. I'm afraid your solution allows to have mixed scenarios, with some fields using placeholders and others not. I would prefer to have placeholders always mandatory or remove them completely. It would be great if we can decide this at config level (i.e. if we want to use placeholders for a project, they are mandatory, if not, it should not be possible to pass them to a field)

gabro:
> I would prefer to have placeholders always mandatory or remove them completely

I don't think this is something we can achieve at the DS level. Let's discuss this better in person when @Phyele is back to work

veej (approved):

PR #466, opened 2022-12-05
Downgrade @react-aria/focus to fix issue with React strict mode

gabro:
react-aria/focus broke compatibility with React strict mode.

See https://github.com/adobe/react-spectrum/issues/3458

3.7.0 is the last version that works, and given we have a very limited use of FocusScope it's fine for us to pin a lower version


veej (approved):
👍

PR #467, opened 2022-12-06
Add missing padding in Table and column cells

gabro:
We were missing a few paddings wrt the design, specifically:
- the horizontal padding of 8px around the whole table
- the padding around Button, ButtonLink and IconButton cells

### DIFF
--- a/packages/bento-design-system/src/Table/cells.tsx
+++ b/packages/bento-design-system/src/Table/cells.tsx
@@ -31,9 +31,11 @@ export function ButtonCell({
   options: Partial<Pick<ButtonProps, "size">>;
 }) {
   return (
-    <Inline space={0} align={align} alignY="center">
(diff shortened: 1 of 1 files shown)


packages/bento-design-system/src/Table/Table.tsx
```diff
@@ -387,11 +394,11 @@ function ColumnHeader<D extends Record<string, unknown>>({
 
   return (
     <Box
-      className={clsx(lastLeftSticky && lastLeftStickyColumn, stickyHeaders && stickyColumnHeader)}
+      className={[lastLeftSticky && lastLeftStickyColumn, stickyHeaders && stickyColumnHeader]}
```
gabro, line 397: not strictly related, but `clsx` is useless here since `Box` already uses it internally to merge the classes

veej (approved):
Looks 💯

PR #474, opened 2022-12-15
Show rightAccessory in ReadOnlyField

### LINKED ISSUE #473: Bug: ReadOnlyField ignores rightAccessory
### Description of the bug

ReadOnlyField accepts a rightAccessory prop it gets ignored, I would assume it should ignore it only if I use the copyButton props and not always.

To reproduce just use the ReadOnlyField component and pass a rightAccessory with any content, it won't be rendered.

### Playroom reproduction

https://playroom.bento.buildo.io/#?code=N4Igxg9gJgpiBcIA8AlGBDKB5AdgGwE8AxASxjygAIA3dPAVxgF4AdEWhmNyvdAI3KsQpclG4AnEgHMAFgBcAgmDAwAzqojiCQsPVVyIAW24B6AHwscSACowAHnJEUadRkI6NuvAXiFOxIJSSsorKahpaOnoGxoHmliAANCAA7iRQcjKqCADaAOwAbAAcALoAvkA

### Version

latest

_No response_

gabro:
Fixes #473

It was previously ignored, due to a bug.

This also makes sure we still display the `rightAccessory` in case there is already a copy button

### DIFF
--- a/packages/bento-design-system/src/ReadOnlyField/ReadOnlyField.tsx
+++ b/packages/bento-design-system/src/ReadOnlyField/ReadOnlyField.tsx
@@ -1,5 +1,7 @@
+import { match, __, not } from "ts-pattern";
 import { useBentoConfig } from "../BentoConfigContext";
 import { IconButton } from "../IconButton/IconButton";
+import { Columns } from "../Layout/Columns";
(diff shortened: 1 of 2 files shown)


veej (approved):
Looks :100

PR #476, opened 2022-12-15
#472: Add support for units of measurement on BarChart component

### LINKED ISSUE #472: Add support for units of measurement on BarChart component
### Discussed in https://github.com/buildo/bento-design-system/discussions/470

federico-ercoles:
Closes #472

The request was only to implement support for units of measurement. This proposed solution allows users to define a generic formatter for values on both X and Y axis, to avoid having to differentiate between pre- and post-fixed units.
~~Additionally, this solution is only implemented for the `BarChart` component, since that was the request, but can easily be extended to the `DonutChart` (where it may be only relevant for tooltips) and `LineChart` components.~~

### DIFF
--- a/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
+++ b/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
@@ -13,7 +13,8 @@ import { allColors } from "../../util/atoms";
 import { vars } from "../../vars.css";
 import { ChartProps } from "../ChartProps";
 import { legendContent } from "../Legend/Legend";
-import { tooltipContent } from "../Tooltip/Tooltip";
+import { TooltipContent } from "../Tooltip/Tooltip";
+import { ValueFormatter, defaultValueFormatter, ValueType, NameType } from "../utils";
 
 type Props<D extends string, C extends string> = ChartProps & {
   data: Record<D | C, unknown>[];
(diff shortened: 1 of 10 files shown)


packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
```diff
@@ -45,6 +48,8 @@ export function BarChart<D extends string, C extends string>({
   stacked = false,
   dataColors,
   children,
+  xAxisValueFormatter = defaultValueFormatter,
+  yAxisValueFormatter = defaultValueFormatter,
```
gabro, line 52: why do we need a `defaultValueFormatter`? Can't we just pass `undefined`?

packages/bento-design-system/src/Charts/Tooltip/Tooltip.tsx
```diff
@@ -3,15 +3,21 @@ import { Box } from "../../Box/Box";
...
+import { NameType, ValueFormatter, ValueType } from "../utils";
 
-type ValueType = number | string;
-type NameType = number | string;
+type TooltipProps<TValue extends ValueType, TName extends NameType> = RechartsTooltipProps<
+  TValue,
+  TName
+> & {
+  valueFormatter?: ValueFormatter;
+};
```
gabro, line 13: I think we can use the standard `formatter` value from `Tooltip`. If we do that we can avoid this intermediate `TooltipComponent` wrapping and keep using `tooltipComponent` as before, and passing `yAxisValueFormatter` to `Tooltip` directly.

I have a working prototype (based on this branch) here: https://github.com/buildo/bento-design-system/compare/472-add_support_for_units-2?expand=1

Let me know what you think

gabro (changes requested):
See the inline comments.

Additionally, let's add it wherever makes sense (so also LineChart and DonutChart)

veej (approved):

gabro (approved):

PR #482, opened 2022-12-19
#477: Make charts Tooltip follow Card style

### LINKED ISSUE #477: Make charts tooltip customizable
### Discussed in https://github.com/buildo/bento-design-system/discussions/471

<div type='discussions-op-text'>

<sup>Originally posted by **stefanosambruna** December 15, 2022</sup>
We would like to customize graphs tooltip in this way but at the moment I guess it's not possible

image
</div>

federico-ercoles:
Closes #477

### DIFF
--- a/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
+++ b/packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
@@ -13,7 +13,7 @@ import { allColors } from "../../util/atoms";
 import { vars } from "../../vars.css";
 import { ChartProps } from "../ChartProps";
 import { legendContent } from "../Legend/Legend";
-import { tooltipContent } from "../Tooltip/Tooltip";
+import { tooltipContent, tooltipStyle } from "../Tooltip/Tooltip";
 import { ValueFormatter } from "../ValueFormatter";
 
 type Props<D extends string, C extends string> = ChartProps & {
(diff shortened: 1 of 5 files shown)


packages/bento-design-system/src/Charts/Tooltip/Tooltip.tsx
```diff
@@ -1,51 +1,66 @@
...
-            return (
-              <Columns key={name} space={4} alignY="center">
-                <Column width="content">
-                  <Box height={16} width={16} borderRadius={4} style={{ backgroundColor: color }} />
-                </Column>
-                <Body size="small">{formattedText}</Body>
-              </Columns>
-            );
-          })}
+export const useTooltipContent = () => {
```
federico-ercoles, line 10: I had to change this to a hook in order to call `useBentoConfig` to retrieve the config value for the border radius.

packages/bento-design-system/src/Charts/BarChart/BarChart.tsx
```diff
@@ -52,6 +52,7 @@ export function BarChart<D extends string, C extends string>({
   yAxisValueFormatter,
 }: Props<D, C>) {
   const config = useBentoConfig();
+  const { tooltipContent, tooltipStyle } = useTooltip();
```
veej, line 55: consider to return directly the Tooltip component here, so that we don't need to remember to pass the `wrapperStyle`

gabro (approved):

PR #499, opened 2023-01-13
Add spacing token 12

federico-ercoles:


gabro (approved):
Can you check if we mention the spacing scale explicitly in the documentation and update it if that's the case?

PR #501, opened 2023-01-13
Input background

federico-ercoles:
Default background for `NumberInput` as `brandPrimary`:
Screenshot 2023-01-13 at 14 51 46

Read-only background for `NumberInput` as `brandTertiary`:
Screenshot 2023-01-13 at 14 51 55

### DIFF
--- a/packages/bento-design-system/src/Field/utils.ts
+++ b/packages/bento-design-system/src/Field/utils.ts
@@ -1,0 +1,12 @@
+import { assignInlineVars } from "@vanilla-extract/dynamic";
+import { statusProperties } from "../util/atoms";
+import { InputConfig } from "./Config";
+import { readOnlyBackground } from "./Field.css";
+import { normalizeStatusValue } from "../internal/sprinkles.css";
+
+export function getReadOnlyBackgroundStyle<T extends InputConfig>(config: T) {
+  return assignInlineVars({
+    [readOnlyBackground]:
+      statusProperties.background[normalizeStatusValue(config.background.readOnly).default!],
+  });
+}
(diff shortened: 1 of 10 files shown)


packages/bento-design-system/src/Field/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { assignInlineVars } from "@vanilla-extract/dynamic";
+import { statusProperties } from "../util/atoms";
+import { InputConfig } from "./Config";
+import { readOnlyBackground } from "./Field.css";
+import { normalizeStatusValue } from "../internal/sprinkles.css";
+
+export function getReadOnlyBackgroundStyle<T extends InputConfig>(config: T) {
+  return assignInlineVars({
+    [readOnlyBackground]:
+      statusProperties.background[normalizeStatusValue(config.background.readOnly).default!],
```
gabro, line 10: maybe we can remove the `!` by using `NonNullable` in the type definition? Or, actually, is it needed at all?

gabro (commented):
A tiny comment

packages/bento-design-system/src/Field/utils.ts
federico-ercoles: It's needed because in the object returned from `normalizeStatusValue` `default` is an optional field, but, from what I understand of how it should work, I think it should always be defined? Not really sure.
Wrapping the type definition `BentoSprinkes["background"]` with `NonNullable` didn't fix the issue.

gabro: ah, wait, we don't need `normalizeStatusValue` anymore, since we're using `& string` in the definition. Have you tried just:

```suggestion
      statusProperties.background[config.background.readOnly],
```

?

gabro (approved):

PR #511, opened 2023-01-18
Allow separate config of paddingX and paddingY for IconButton

federico-ercoles:


packages/bento-design-system/src/IconButton/Config.ts
```diff
@@ -3,5 +3,8 @@ import { BentoSprinkles } from "../internal";
 
 export type IconButtonConfig = {
   radius: BentoSprinkles["borderRadius"];
-  padding: Record<IconProps["size"], BentoSprinkles["padding"]>;
+  padding: Record<
+    IconProps["size"],
+    { paddingX: BentoSprinkles["paddingX"]; paddingY: BentoSprinkles["paddingY"] }
+  >;
```
gabro, line 9: we could try making it non-breaking, by accepting also a single padding:

```suggestion
  padding: Record<
    IconProps["size"],
    BentoSprinkles["padding"] | { paddingX: BentoSprinkles["paddingX"]; paddingY: BentoSprinkles["paddingY"] }
  >;
```

gabro (changes requested):

gabro (approved):

PR #529, opened 2023-02-22
Add support for ESM

gabro:
Adds support for ESM to Bento, which enables significantly better tree-shaking.

We successfully tested this in a real-world project and observed an important bundle reduction.

While working on this, I had to change some things slightly; I'll add some comments inline.

I've marked the PR as breaking since if the consumer is using an ESM-enabled bundler (very likely), they will need to upgrade their imports from `/lib/...` to `/...` (e.g., `import "@buildo/bento-design-system/lib/index.css"` -> `import "@buildo/bento-design-system/index.css"`)


veej (approved):
Looks 💯

PR #534, opened 2023-02-26
fix(docs): Fix typos in Quick Start > Project Structure section

arabello:
Fix a couple of typos in the `Quick Start` section of the documentation:
-  `noExternals` => `noExternal` as the correct option is without the `s`
- `index.tsx` => `index.ts` as there is no JSX syntax in the index file. This is debatable, but there is inconsistency across the page: if we want to keep `index.tsx`, an occurrence of `index.ts` should be updated.

# Test Plan
Run the website locally
Screen Shot 2023-02-26 at 22 04 58
Screen Shot 2023-02-26 at 22 05 14

### DIFF
--- a/packages/website/docs/02-Getting Started/03-project-structure.mdx
+++ b/packages/website/docs/02-Getting Started/03-project-structure.mdx
@@ -102,7 +102,7 @@ export default defineConfig({
   inject: ["./jsxShim.ts"],
   // Include here the css files coming from external dependencies, which we
   // recommend to bundle in your design system package.
-  noExternals: [
+  noExternal: [
     "@buildo/bento-design-system/lib/index.css",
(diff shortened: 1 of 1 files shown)


packages/website/docs/02-Getting Started/03-project-structure.mdx
```diff
@@ -147,11 +147,11 @@ my-project/
 ├─ pnpm-workspaces.yaml
 ```
 
-`packages/design-system/src/index.tsx` is where you configure and export Bento components.
+`packages/design-system/src/index.ts` is where you configure and export Bento components.
 
 Your `index.ts` will look something like this:
```
arabello, line 152: The occurrence of `index.ts` to updated if we prefer using `.tsx` extension

gabro (approved):
LGTM! Thanks

Issue #540, opened 2023-03-07
Bug: Build time dts regression


gffuma:
### Description of the bug

Switching to Bento >= 0.15.5 make the build time of dts changes from ~2 s to ~24 s (on my MacBookPro M2 2022).

I follow the instructions of design system repository from the documentation.
I create a repository to reproduce the bug:

Fast:
https://github.com/gffuma/repro-bento-build-time-regression

Slow:
https://github.com/gffuma/repro-bento-build-time-regression/tree/slow

### Version

>= 0.15.5

gabro:
Closed by #542

PR #542, opened 2023-03-08
Downgrade sprinkles to v1.4.1 (to solve #540)

gabro:
Temporary fix for #540.

This essentially reverts commit 076b6633cae4eb2d14aff0c0d3dfca2cf65b2572.

I've published a beta version 0.17.1-beta.1, tested against the repo in #540, and now I see the .d.ts being generated in a reasonable amount of time (~5s on my machine)


veej (approved):
💯

PR #548, opened 2023-03-13
Better SSR support

gabro:
To use Bento in an SSR environment, the consumer needs to wrap the app in `SSRProvider` from react-aria. However, react-aria is an implementation detail, and we don't document this anywhere.

This PR:
- re-exports `SSRProvider` from react-aria
- adds a `ssr` prop to `BentoProvider` to that the provider can be applied automatically (without even knowing about react-aria)
- adds some docs about this

### DIFF
--- a/packages/bento-design-system/src/index.ts
+++ b/packages/bento-design-system/src/index.ts
@@ -119,3 +119,6 @@ export { useComponentsShowcase } from "./useComponentsShowcase";
 export type { ClassValue } from "clsx";
 
 export { Time } from "@internationalized/date";
+
+export { SSRProvider } from "@react-aria/ssr";
+export type { SSRProviderProps } from "@react-aria/ssr";
--- a/packages/bento-design-system/package.json
+++ b/packages/bento-design-system/package.json
@@ -75,6 +75,7 @@
     "@react-aria/radio": "3.4.0",
     "@react-aria/separator": "3.2.4",
     "@react-aria/slider": "3.2.2",
+    "@react-aria/ssr": "3.5.0",
     "@react-aria/switch": "3.2.4",
     "@react-aria/textfield": "3.7.2",
     "@react-aria/tooltip": "3.3.2",
--- a/packages/bento-design-system/src/BentoProvider.tsx
+++ b/packages/bento-design-system/src/BentoProvider.tsx
@@ -1,4 +1,4 @@
-import { Children, PartialBentoConfig } from ".";
+import { Children, PartialBentoConfig, SSRProvider } from ".";
 import { bentoSprinkles } from "./internal";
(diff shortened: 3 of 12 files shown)


packages/bento-design-system/src/index.ts
```diff
@@ -119,3 +119,6 @@ export { useComponentsShowcase } from "./useComponentsShowcase";
 export type { ClassValue } from "clsx";
 
 export { Time } from "@internationalized/date";
+
+export { SSRProvider } from "@react-aria/ssr";
+export type { SSRProviderProps } from "@react-aria/ssr";
```
gabro, line 124: This is the same approach as react-spectrum. I don't think we need necessarily need this export, but in case someone composes their provider differently, I think it makes sense to expose it.

bvkimball:
@gabro bento also needs a check to see if it is SSR so that it knows to skip any reference to document, which can be found anywhere a portal is created. 

https://github.com/buildo/bento-design-system/blob/8b2413a46ec61ecd88984e845bbd170e778384f2/packages/bento-design-system/src/util/createPortal.ts#L6

https://github.com/buildo/bento-design-system/blob/8b2413a46ec61ecd88984e845bbd170e778384f2/packages/bento-design-system/src/Menu/NestedMenu.tsx#L76

https://github.com/buildo/bento-design-system/blob/8b2413a46ec61ecd88984e845bbd170e778384f2/packages/bento-design-system/src/SelectField/SelectField.tsx#L101

https://github.com/buildo/bento-design-system/blob/8b2413a46ec61ecd88984e845bbd170e778384f2/packages/bento-design-system/src/DateField/Day.tsx#L95

https://github.com/buildo/bento-design-system/blob/8b2413a46ec61ecd88984e845bbd170e778384f2/packages/bento-design-system/src/Table/Table.tsx#L207

I think this is just for the docs but for completeness
https://github.com/buildo/bento-design-system/blob/8b2413a46ec61ecd88984e845bbd170e778384f2/packages/bento-design-system/src/useComponentsShowcase.tsx#L64

gabro:
Thank you @bvkimball for spotting these! I'm not super familiar with SSR myself, but it seems most of them are inside useEffect, which should only execute client-side, so they should be ok, does it sound right to you?

I'll take a look at the ones that are not and figure out a strategy.

bvkimball:
@gabro Yes that does sound right, I have only gotten the error when using "SelectField" because of the useMemo. I will confirm, I did not see it anywhere else.  I was just looking for the use of document.

gabro:
@bvkimball ok, thanks for confirming, I've updated the PR to fix the outstanding cases. I'll take a second pass tomorrow and merge it

veej (approved):
Just a minor doubt, but 👍

Issue #549, opened 2023-03-14
Bug: Missing needed native HTML properties 


bvkimball:
### Description of the bug

A button can have a `name` and a `value` for when it is used within a Form.  You can set these native HTMLButtonElements because of the typings.

I think you should include something like the following

```diff
type Props = {
  label: LocalizedString;
 // ...
} 
& Omit<AriaButtonProps<"button">, "onPress"> 
+ & React.HTMLProps<HTMLButtonElement>;
```

or Just the props you want to expose include
```diff
+ & Pick<React.HTMLProps<HTMLButtonElement>, "name" | "id">;
```

or  cleaner 
```diff
+ type IncludedHTMLButtonProps = Pick<React.HTMLProps<HTMLButtonElement>, "name" | "id">;
// ...
+ & IncludedHTMLButtonProps;
```

I think you should do the same for TextFields too, I personally need access to `onKeyDown` event

```diff
+ & React.HTMLProps<HTMLInputElement>;
```

### Playroom reproduction

https://playroom.bento.buildo.io/#?code=N4Igxg9gJgpiBcIA8AhArgFwxAdgAgBsBDAIxgIF4AdEAFRgGcMa8ciBbGakIsDAS1wsA9AD4qOEAF8gA

### Version

1.17.1

### What browsers are you seeing the problem on?

Chrome

bvkimball:
@gabro I can submit a PR for this.  It might more sense to just add and expose these behaviors in the same DX your already are for additional events..

gabro:
Thank you for reporting @bvkimball! I agree we should support more stuff. If you have time to send a PR I'll gladly review it. As a general rule we tend to start small in terms of API surface and grow it if needed, so I like the cherry-picking approach you proposed.

PR #551, opened 2023-03-15
Add more form-related props to `Button`. Add onKeyDown and onKeyUp to `TextField` and `TextArea`.

bvkimball:
For better native HTML support with Forms
- Added form based attribute to Button like: form, formAction, name and value
- Added onKeyDown, and onKeyUp event to TextField and TextAreaField for key capturing like on "Enter"

To support this I added a new `utils` for Button to get the allowed props we want to include for Button specifically.  This is needed because `useButton` from `react-aria` only returns props it cares about so all others get filtered out.  This util could probably move to the root utils directory if other components start using a similar pattern.

Also added a fix to the Eye and EyeClosed icons to replace `fill-rule` and `clip-rule` with `fillRule` and `clipRule` which React complains about.

This address concern in #549


packages/bento-design-system/src/Button/Button.tsx
```diff
@@ -40,6 +42,7 @@ export function Button(props: Props) {
     // @ts-expect-error
     internal_unsafe__bypassUsePress,
   } = props;
+  const { formButtonProps } = useFormButtonProps(props);
```
gabro, line 45: I see the logic behind this, but I wonder if it's not more practical to add something like `lodash.pick` (1kb gzipped) and turn this whole thing into:

```ts
const otherProps = pick(props, [
 	"name",
  "value",
  "form",
  "formAction",
  "formEncType",
  "formMethod",
  "formNoValidate",
  "formTarget",
	"onKeyDown",
	"onKeyUp"
]);
```

What do you think?

gabro (changes requested):
Thanks! I like the spirit of the PR.

One inline comment

packages/bento-design-system/src/Button/Button.tsx
bvkimball: yeah a little cleaner and keeps everything in Button.tsx,  i have pushed this chagne

packages/bento-design-system/src/Button/Button.tsx
```diff
@@ -8,6 +8,17 @@ import { Label } from "../Typography/Label/Label";
 import { Column, Columns } from "../Layout/Columns";
 import { IconProps } from "..";
 import { useBentoConfig } from "../BentoConfigContext";
+import pick from "lodash.pick";
+
+export type OtherButtonKeys =
```
gabro, line 13: Can we extract the keys to a const and derive the type from it?

Something like

```ts
const otherButtonKeys = [...] as const

type OtherButtonKeys = typeof otherButtonKeys[number]
```

gabro (changes requested):
Nice! One last suggestion so that it's a bit easier to maintain the list of new props

packages/bento-design-system/src/Button/Button.tsx
bvkimball: Pushed, thats a clean pattern i didn't know you could extract those values in TS like that, very useful.

gabro (approved):

PR #554, opened 2023-03-17
Add alternativeNames and relatedComponents to ComponentDoc

gabro:
Example (`TextField`):

image


packages/website/package.json
```diff
@@ -35,7 +35,7 @@
     "assert": "2.0.0",
     "buffer": "6.0.3",
     "clsx": "1.2.1",
-    "docusaurus-plugin-react-docgen-typescript": "1.0.2",
+    "docusaurus-plugin-react-docgen-typescript": "1.1.0",
```
gabro, line 38: this is to fix an issue I've noticed with PropTable not being generated

packages/website/src/snippets/index.tsx
```diff
@@ -1,55 +1,5 @@
...
-  tableColumn,
-  Tabs,
-  TextField,
-  TextArea,
-  Tiles,
-  Toast,
-  Tooltip,
-  SelectField,
-} = createBentoComponents();
+export const DesignSystemProvider = createBentoProvider();
```
gabro, line 5: this was causing funny bugs all around the website; we should stop using `createBentoComponents`

veej (approved):

PR #555, opened 2023-03-17
Generate .d.cts files alongside .d.ts files.

gabro:
This is a workaround for https://github.com/egoist/tsup/issues/579 and more generally for the problems highlighted in this thread https://twitter.com/atcb/status/1634653474041503744?s=61&t=nC5rArtZAYIRm7854OZMDw

Manually checked here using https://arethetypeswrong.github.io/

Here's the result after the current PR

image

Here's how it is currently

image


veej (approved):
👍

PR #565, opened 2023-03-29
#552: Bug: SelectField HTML value not set correctly

### LINKED ISSUE #552: Bug: SelectField HTML value not set correctly
### Description of the bug

Since the value prop/attribute is assigned the entire option object instead of the value it matches, it doesn't seem to be a valid form value.  I am not using the Bento `Form` component since i am doing more traditional HTML workflow with form posts.  I also have not dug into the bento `Form` workflow, so this could be entirely intentional on your part.

Here is the relevant code:
https://github.com/buildo/bento-design-system/blob/main/packages/bento-design-system/src/SelectField/SelectField.tsx#LL136C11-L140C12

My Playroom is a little convoluted because i needed to use a native html form.  but you can see that `color` is never added to the formData because it needs to be a string value.

Sorry for the bug reports, but I really like your component library and I really like how easy it is to override and customize to my style and needs.

Again I can submit a PR, if this is something worth changing, unless there are downstream affects to be considered for this.

### Playroom reproduction

https://playroom.bento.buildo.io/#?code=N4Igxg9gJgpiBcJgApkEoAEBeAfB4AOgHYYaREDOALhgNpECGAtjADQYUxUByzMAutgwAlGAzBUAdAFdOAZSoMqMZARBq0AbmKly1OpAA2EAE7tOVAMIRjJwVhFiJM%2BYuWr1ILcR1kIlGgALBiIoQxg5aQAjJgBLGgdkGExcfF9df30AM1MmABElBiEiGAB3DAAxXILFJMlFEwBzLm8SUj9KGxhJY0bkAHkogCsYZyyTCCYAUSIqE1iYCmQck3zCtFbSAF9tIl8TLmkTElU20gAeFaYMf0iY%2BKxgYNDwu7iqLZx09vOq1blRlRYv4vmd2hc-kxhBBSqDweDzgAVGAADyoFQWhig33hGEYLCwanxMDUOPhhgYURghkJIF4LFJYNxAAcKWAYIEbLATLTEQBPZkwDCBGAHSTixm48EANwYhmkMEexK2ZPB-ksz2ajws9JgKqZ7QA9HCpedDZDobDVRhfrlLSapRhrRcAeEJBjqdiDbjZfLFcAjKZ9Y61UQNSEtcALNZbMGQ6RibTAyZJfGMBSqTS1AB1YJUADkFAwfIgRwwWQY0tM8SFyYA-Kn46zxByuaLaa7ARgisnGyGIMygZlHrRnfDCN6Q76FfAMABGVhj3EZ6mztSibEgReTx0Aa1ioTXICopQgAFpDAeSVul-DOOQoAAZK9HgDEWSyAAZv5%2B1Nu0%2B0Wz-gBaQ7lK04wLOABMwEgemlKrk6IAAEJ%2Bn%2Bt7gvuh5ISe56XiU6Fgbi97%2BE%2BL5Ia%2BP6fh%2BhFwUBGGgXBpAQbOADMsEgSuhhHgA4gcMB7DeRHwlhUBHrhF5XrRTEkaEz4lG%2B34ft%2B0kAfRwn8HGjrGtaZoWjCDo-OauQAhIwJEIZFxRNIVBUP4GBUAKipqBQ0TvGoOBvPEZrWbZILfGaVwmq0WxoOg%2BogFsQA

### Version

1.17.1

Chrome

federico-ercoles:
Closes #552

### tests performed

### DIFF
--- a/packages/bento-design-system/src/SelectField/SelectField.tsx
+++ b/packages/bento-design-system/src/SelectField/SelectField.tsx
@@ -76,6 +76,7 @@ export function SelectField<A>(props: Props<A>) {
     options,
     onBlur,
     label,
+    name,
     hint,
     hintPlacement,
     assistiveText,
@@ -131,6 +132,7 @@ export function SelectField<A>(props: Props<A>) {
       >
         <Select
           id={fieldProps.id}
+          name={name}
           aria-label={fieldProps["aria-label"]}
           aria-labelledby={fieldProps["aria-labelledby"]}
           isDisabled={disabled}


gabro (changes requested):
Changes LGTM, but I don't get the pnpm-lock huge diff. Can you double check?

PR #570, opened 2023-04-06
improve turborepo config

veej:
This PR improves turborepo config by:
* specifying output dirs for all the build tasks, so that we can properly take advantage of the turborepo's cache
* correctly mark development tasks as persistent + disable the cache for them, as suggested by the doc
*  avoiding building all the workspaces when playroom-start/playroom-build scripts are run, due to the turborepo's behaviour described [here](https://github.com/vercel/turbo/issues/937), by targeting exactly the buildo-design-system workspace in the dependsOn property
* add a wait mechanism for the "start" script, so that storybook starts watching only after buildo-design-system has been compiled at least once, as described [here](https://github.com/vercel/turbo/issues/1497#issuecomment-1424852247)

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -5,6 +5,10 @@
     "build": {
       "dependsOn": [
         "^build"
+      ],
+      "outputs": [
+        "lib",
+        "storybook-static"
       ]
     },
     "typecheck": {
@@ -13,7 +17,11 @@
       ]
     },
     "start": {
-      "outputs": []
+      "persistent": true,
+      "cache": false,
+      "dependsOn": [
+        "^wait"
+      ]
     },
     "prettier-check": {},
     "prettier-write": {},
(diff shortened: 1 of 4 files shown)


turbo.json
```diff
@@ -26,21 +34,38 @@
...
+      ],
+      "outputs": [
+        "build"
       ]
     },
-    "check-circular-deps": {}
+    "check-circular-deps": {},
+    "wait": {
+      "cache": false
+    }
```
gabro, line 69: do we need this? Isn't it sufficient to define a `wait` script in the project?

veej: > Isn't it sufficient to define a `wait` script in the project?

nope. It seems scripts used in the `dependsOn` must be defined in `turbo.json`, otherwise we get a:

> Error preparing engine: cannot find task for ___ROOT___#wait

gabro (approved):

PR #580, opened 2023-04-19
Add ThemeProvider

veej:
image

In the story above, the default theme has been overridden inside the modal to change the primary button color.


packages/bento-design-system/src/BentoThemeContext.tsx
```diff
@@ -0,0 +1,35 @@
...
+  // NOTE(vince): in case we nest theme providers, each nested provider should only override the
+  // partial theme it was given.
+  // We may just set the variables we want to override, but this wouldn't work with Portals,
+  // since the base theme would be inherited from the root element and not from the parent theme provider.
+  const parentTheme = useBentoTheme();
+  const theme = deepmerge(parentTheme, value) as BentoTheme;
+
+  return (
+    <BentoThemeContext.Provider value={theme}>
+      <Box style={assignInlineVars(vars, theme)}>{children}</Box>
```
veej, line 32: The only way I found to set the theme dynamically is to create this intermediate Box and set the inline style. This may affect the app layout.

packages/bento-design-system/src/util/useCreatePortal.tsx
```diff
@@ -2,8 +2,13 @@ import { ReactPortal } from "react";
...
 import { Children } from "./Children";
 import { useIsSSR } from "@react-aria/ssr";
+import { BentoThemeProvider, useBentoTheme } from "../BentoThemeContext";
 
 export function useCreatePortal(): (children: Children) => ReactPortal | null {
   const isSSR = useIsSSR();
-  return (children) => (isSSR ? null : createReactPortal(children, document.body));
+  const theme = useBentoTheme();
+  return (children) => {
+    const content = <BentoThemeProvider value={theme}>{children}</BentoThemeProvider>;
```
veej, line 11: we need to wrap the Portal children inside a new `BentoThemeProvider`, otherwise the Portal content won't inherit the style from the Box we add with `BentoThemeProvider`

veej (commented):
@gabro  I'm not really happy with this solution, as it requires to add intermediate divs that may break the app layout. The inline variables are also unhandy to modify externally, as we won't have a style set on the :root element, but divs scattered around the app's tree which may be uneasy to target

packages/bento-design-system/src/defaultThemeClass.css.ts
```diff
@@ -0,0 +1,5 @@
+import { createTheme } from "@vanilla-extract/css";
```
veej, line 1: the `defaultTheme.css.ts` above is exporting the default theme as global, so that the user can just import `defaultTheme.css` and the style will be attached to the `:root` element.
If the user wants to use the `BentoThemeProvider` instead (i.e. creating a div with the theme className applied, for example to allow switching between different themes in the app), we also export the defaultTheme as a class (with `createTheme` instead of `createGlobalTheme`).

packages/bento-design-system/src/util/makeBentoTheme.ts
```diff
@@ -0,0 +1,13 @@
+import { deepmerge } from "deepmerge-ts";
+import { PartialDeep } from "./PartialDeep";
+import { BentoTokens } from "./bentoTokens";
+import { defaultTokens } from "./defaultTokens";
+import { createTheme } from "@vanilla-extract/css";
+import { vars } from "../vars.css";
+
+export function makeBentoTheme(tokens: PartialDeep<BentoTokens>) {
```
veej, line 8: With this utility, users can create new themes easily by partially overriding the bento default theme. If this utility is used, we don't have to worry about adding new breaking tokens

veej:
Not sure why all the Chromatic screenshots have an additional padding now. Stories look exactly the same as before

veej:
Added also a support to inline theming and theme Box customization via `as` and `className` properties.

It's now possible to use the `BentoThemeProvider` to override tokens inline, so for example we can:

1. Override the whole theme with another static theme created with `makeBentoTheme`:
```
<BentoThemeProvider theme={{ theme: newTheme }}>
```

2. Override just a few tokens from the parent theme, with:
```
<BentoThemeProvider theme={{ tokenOverrides: { interactiveBackgroundColor: { primarySolidFocusBackground: "red" } } }}>
```

3. Do both, passing both the theme and the `tokenOverrides` fields.

Note that `tokenOverrides` can be set progressively, as they're merged with the overrides defined in the parent provider.

packages/bento-design-system/src/BentoProvider.tsx
```diff
@@ -37,12 +38,14 @@ type Props = {
...
   locale?: string;
   config?: PartialBentoConfig;
+  theme?: BentoTheme | null;
   sprinkles?: SprinklesFn;
   ssr?: boolean;
 } & DefaultMessages;
 
 export function createBentoProvider(
   config: PartialBentoConfig = {},
+  theme?: BentoTheme | null,
```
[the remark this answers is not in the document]
gabro: this is technically breaking, isn't it? I mean, if you were using `createBentoProvider` + global static themes, now you have to explicitly pass `null` here to avoid duplicating everything.

Also, I think exploiting the difference between `null` and `undefined` may be a bit cryptical, maybe we can be more explicit?

packages/bento-design-system/src/BentoProvider.tsx
```diff
@@ -38,16 +38,28 @@ type Props = {
...
 } & DefaultMessages;
 
 export function createBentoProvider(
   config: PartialBentoConfig = {},
-  theme?: BentoThemeOverride,
+  theme?: BentoTheme | null,
   sprinkles: SprinklesFn = bentoSprinkles
 ) {
+  function OptionalThemeWrapper(props: { children: Children; theme?: BentoTheme | null }) {
+    if (theme === null) return <>{props.children}</>;
```
gabro, line 58:
```suggestion
    if (props.theme === null) return <>{props.children}</>;
```

?

packages/bento-design-system/src/BentoProvider.tsx
```diff
@@ -38,16 +38,28 @@ type Props = {
...
+    const _theme = props.theme ?? defaultTheme;
+    return <BentoThemeProvider theme={_theme || defaultTheme}>{props.children}</BentoThemeProvider>;
```
gabro, line 60: I'm a bit confused by all the defaulting here. Why `_theme || defaultTheme` if  `_theme` is already falling back to `defaultTheme`. Maybe just:

```suggestion
    const _theme = props.theme ?? defaultTheme;
    return <BentoThemeProvider theme={_theme}>{props.children}</BentoThemeProvider>;
```

?

gabro (changes requested):
a few comments, I will need to re-review this later today

packages/bento-design-system/src/BentoProvider.tsx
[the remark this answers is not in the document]
veej: I agree on the second part. I'm not happy with it either. I thought of passing a special value, like `global` instead of `null`, but the type-checking would be weak given that the theme className is also a string. 
I'm open to suggestions

gabro (approved):

PR #592, opened 2023-05-11
Avoid merging JSX.Elements in config

veej:
**Before:**
We're correctly using the innermost configuration for the activeElement background color, but we're also inheriting the `borderRadius` from the outermost configuration.
image

**After:**
We're replacing the activeElement completely, so we're no longer inheriting the borderRadius from the old activeElement config. We're still inheriting other configs, like `uppercaseLabel`.
image


gabro (approved):
Nice!

PR #597, opened 2023-05-16
Changes to the default theme, config and icons to align them with Figma v2.7

veej:


gabro (approved):
Code is 👍 but we need to document all of these changes in the release notes. Specifically:
- all the icons renames
- default config renames

The ideal thing would be that a user has a clear path to upgrade while maintaining the UI unchanged

PR #598, opened 2023-05-17
New feedback component

veej:
image

image

image


packages/bento-design-system/src/util/defaultConfigs.tsx
```diff
@@ -204,31 +204,30 @@ export const disclosureGroup: DisclosureGroupConfig = {
...
 export const feedback: FeedbackConfig = {
-  background: null,
-  positiveIllustration: IllustrationPositive,
-  negativeIllustration: IllustrationNegative,
+  positiveIcon: IconPositiveCircle,
+  negativeIcon: IconNegative,
   title: {
     medium: { kind: "title", size: "large" },
-    large: { kind: "display", size: "small" },
+    large: { kind: "headline", size: "medium" },
```
gabro, line 211: Let's document the changes in this file in the migration guide: if the user wants to preserve the same UI, they need to manually set the config to what it was before.
I would provide a code snippet to add to their config.

gabro (approved):
LGTM, let's document the migration guide well in the release notes

PR #599, opened 2023-05-17
Remove illustrations

veej:


gabro (approved):
Changes look 💯.
Let's make sure to document how to migrate off using illustrations in the release notes.
What comes to mind:
- if you were using illustrations directly: download the SVG from Figma and turn it into a React component
- if you were passing illustrations to lists or navigation items: you can work around it by passing the old illustration to the `icon` prop, as in `icon={() => <MyIllustration size={24} kind="outline" />`
- if you

PR #617, opened 2023-06-04
Allow setting different label sizes per button size

### LINKED ISSUE #616: Bug: Button Labels don't scale with size
### Description of the bug

Button text labels have a single fixed size while buttons can scale L-M-S.

Compare to MUI https://mui.com/material-ui/react-button/#sizes where the label text scales appropriately.

It is impossible to select an appropriate text size for all three sizes

### Version

2.7

### Figma version

Latest

gabro:
Fixes #616 

This allows configuring a variable Button label size based on the button size.

This is not breaking since the `defaultConfig` stays the same


veej (approved):
💯

PR #623, opened 2023-06-08
Migrate to Storybook 7

gabro:
Kind of a huge PR, but fairly mechanical. Let's see what Chromatic has to say


packages/storybook/.storybook/main.ts
```diff
@@ -0,0 +1,33 @@
...
+  stories: [{ directory: "../stories", files: "**/*.stories.*" }],
+  addons: ["@storybook/addon-links", "@storybook/addon-essentials", "storybook-addon-themes"],
+  framework: {
+    name: "@storybook/react-webpack5",
+    options: {},
+  },
+  webpackFinal: (baseConfig) => {
+    const { module: { rules = [] } = {}, plugins = [] } = baseConfig;
+
+    // by default, the rules for jsx|tsx files are including the whole monorepo root,
```
veej, line 17: is this still true? Does it make sense to check if we can remove these lines?

packages/storybook/.storybook/preview.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const onChange = (value: unknown) => {
+      (ctx.args.onChange as (value: unknown) => void)?.(value);
+
+      // Check if the component is controlled
+      if (ctx.args.value !== undefined) {
+        setArgs({ value });
+      }
+    };
+
+    return <Story args={{ ...ctx.args, onChange }} />;
```
veej, line 27: Couldn't be useful to allow to override the `onChange` via args?

veej (approved):
Looks good, even if I didn't check all the stories. Just a couple of minor comments inline

packages/storybook/.storybook/preview.tsx
gabro: maybe, I'll change it

gabro: done

packages/storybook/.storybook/main.ts
gabro: I haven't seen anything specific to the webpack config in the changelog, so I'll keep it for now

PR #632, opened 2023-06-23
Improve doc for LocalizedString + i18next

veej:


packages/website/docs/06-Advanced/strict-localization.mdx
```diff
@@ -97,32 +97,41 @@ export function useFormatMessage(): (
...
+import enMessages from "./locales/en.json";
+
+type Primitive = string | boolean | number | null | undefined;
+type MapLeafNodes<Obj, LeafType> = {
+  [Prop in keyof Obj]: Obj[Prop] extends Primitive
+    ? LeafType
+    : Obj[Prop] extends Record<string | number, any>
+    ? MapLeafNodes<Obj[Prop], LeafType>
+    : never;
+};
```
gabro, line 110: By the way, this MapLeafNode works well for an arbitrarily nested structure, but aren't translation strings flat? Maybe this could be a simpler mapped type?

gabro (approved):
❤️ just a comment, but already good

packages/website/docs/06-Advanced/strict-localization.mdx
veej: Doesn't `i18next` support nested keys as `react-intl` does? e.g.

```
{ 
   "MyView": {
       "MySection": {
          "key1": "Key 1",
          ...
       }
   }
}
```

Maybe we can simplify the `Primitive` type and check only for strings, but I think the nesting is useful

gabro: I didn't remember that was allowed. Let's leave it like this then

veej: yep, it is:
https://www.i18next.com/misc/json-format -> see the "keyDeep"

PR #636, opened 2023-06-26
Replace jest with vitest

gabro:
One less thing that uses babel 🎉


veej (commented):
Looks good, but it seems the CI is running vitest in watch mode. Could you check?

gabro:
Oh wops. I'll fix it

veej (approved):

PR #637, opened 2023-06-26
Upgrade tsup and remove patch-esm-ts script

gabro:
tsup 7.1.0 introduced [this change](https://github.com/egoist/tsup/pull/934), so we no longer need our custom workaround.

Eventually I would like to also introduce a CI step which uses https://github.com/arethetypeswrong/arethetypeswrong.github.io/ (maybe when it's released as a GitHub Action)

For now, I've tested it manually: tsup now generate .d.ts and .d.mts files to match the .mjs and .js files in the bundle.


veej (approved):
Not sure what's wrong in the diff spotted by chromatic, but the PR looks good 👍

PR #645, opened 2023-06-28
Drop React 17

gabro:


veej (approved):
Looks 👍 
Does it make sense to add a few lines in the changelog to explain the rationale behind this choice?

PR #682, opened 2023-07-21
Add defaultElevation and defaultPadding config to Card

federico-ercoles:
Closes BEN-114

### DIFF
--- a/packages/bento-design-system/src/Card/Card.tsx
+++ b/packages/bento-design-system/src/Card/Card.tsx
@@ -3,6 +3,7 @@ import { Box, Children } from "..";
 import { cardRecipe } from "./Card.css";
 import { useBentoConfig } from "../BentoConfigContext";
 import { BorderRadiusConfig, getRadiusPropsFromConfig } from "../util/BorderRadiusConfig";
+import { CardElevation } from "./Config";
 
 type PaddingKey =
   | "padding"
@@ -15,7 +16,7 @@ type PaddingKey =
 
 type Props = {
   children: Children;
-  elevation?: "small" | "medium" | "large";
+  elevation?: CardElevation;
   borderRadius?: BorderRadiusConfig;
 } & { [k in PaddingKey]?: 0 | BentoSprinkles["padding"] };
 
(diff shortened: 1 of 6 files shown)


packages/bento-design-system/src/Card/Card.tsx
```diff
@@ -24,7 +25,8 @@ export function Card({ children, elevation, borderRadius, ...boxProps }: Props)
   return (
     <Box
       {...getRadiusPropsFromConfig(borderRadius ?? config.defaultRadius)}
-      className={cardRecipe({ elevation: elevation || "none" })}
+      className={cardRecipe({ elevation: elevation ?? config.defaultElevation })}
```
veej, line 28: I think we're preventing the elevation from being set to none in this way.
Before, it was possible to omit the "elevation" prop and get "none" here as a default value.
In the new implementation, there's no way to get the same behavior.
I think we should add "none" as a possible value in `CardElevation`, so that it can explicitly be passed as prop or be set as the `defaultElevation`.

packages/bento-design-system/src/util/PaddingConfig.ts
```diff
@@ -0,0 +1,13 @@
...
+  | "padding"
+  | "paddingX"
+  | "paddingY"
+  | "paddingLeft"
+  | "paddingRight"
+  | "paddingTop"
+  | "paddingBottom";
+type Padding = 0 | BentoSprinkles["padding"];
+
+export type PaddingConfig = { [k in PaddingKey]?: Padding };
```
veej, line 13: Wouldn't be better to use a more precise union type here, as we did for `BorderRadiusConfig`?
Just to avoid cases like `{ paddingX: 8, paddingLeft: 4, paddingRight: 4 }`.

Also, being able to set `defaultPadding: 4` instead of `defaultPadding: { padding: 4 }` feels more natural to me.

veej (changes requested):
See my comment about elevation: "none" and my doubt about the padding config

veej:
@federico-ercoles have a look also at the diffs on chromatic. I'm not sure they all make sense.

federico-ercoles:
> @federico-ercoles have a look also at the diffs on chromatic. I'm not sure they all make sense.

I'm a bit confused by the diffs on the texts, I'm not finding any difference in the CSS classes applied.

packages/bento-design-system/src/util/PaddingConfig.ts
veej: Didn't understand the first point. Would should it be breaking? Isn't it an additional config entry?
Not sure about the second point either... it's true for sure, but I would prefer to have consistency between configs and avoid nonsense configs instead of optimizing something you have to write only once per project.

veej (approved):
🎉

PR #696, opened 2023-08-11
Reimplement Datepicker with react-aria

veej:
https://github.com/buildo/bento-design-system/assets/925635/ad03c540-9b67-4304-a492-425593d05583


packages/bento-design-system/src/DateField/Calendar.tsx
```diff
@@ -99,8 +99,10 @@ export function Calendar(props: Props) {
           {weekdayLabels.map((d, index) => (
             <Box
               className={weekDay}
-              width={config.dayWidth}
-              height={config.dayHeight}
+              style={{
+                width: config.dayWidth,
+                height: config.dayHeight,
```
veej, line 104: We're no longer limiting the day size in config to be part of the spacing scale.

packages/bento-design-system/src/DateField/CalendarHeader.tsx
```diff
@@ -1,56 +1,54 @@
...
   return (
-    <Box paddingX={12} width="full">
-      <Columns space={4} alignY="center">
+    <Box paddingX={8} width="full">
+      <Columns space={8} alignY="center">
         <Column width="content">
           <IconButton
             label={defaultMessages.DateField.previousMonthLabel}
-            size={16}
+            size={24}
```
veej, line 27: changes already discussed with @Phyele

packages/bento-design-system/src/DateField/DateField.tsx
```diff
@@ -25,104 +23,125 @@ type SingleDateFieldProps = {
 type RangeDateFieldProps = {
   type: "range";
   shortcuts?: ShortcutProps<[Date, Date]>[];
-} & FieldProps<[Date | null, Date | null]>;
+} & FieldProps<[Date, Date] | null>;
```
gabro, line 26: this is breaking and I would like to understand the impact: previously `value` and `onChange` could deal with partial ranges (i.e. either start or end date could be null). Now the types imply that we only deal with complete ranges (either the range is fully defined, or it's `null`).

What are the functional implications: what happens if we set the start date and then leave the field? Does it get erased?

packages/bento-design-system/src/DateField/DateField.tsx
```diff
@@ -25,104 +23,125 @@ type SingleDateFieldProps = {
...
 type Props = (SingleDateFieldProps | RangeDateFieldProps) & {
-  minDate?: UseDatepickerProps["minBookingDate"];
-  maxDate?: UseDatepickerProps["maxBookingDate"];
-  shouldDisableDate?: UseDatepickerProps["isDateBlocked"];
+  minDate?: Date;
+  maxDate?: Date;
+  shouldDisableDate?: (date: DateValue) => boolean;
```
gabro, line 30: what's `DateValue` here? This was previously `(date: Date) => boolean` (looking at `@datepicker-react/hooks` types), so it's potentially breaking.

Can we find a way to keep it as it was?

packages/bento-design-system/src/DateField/DateField.tsx
```diff
@@ -25,104 +23,125 @@ type SingleDateFieldProps = {
...
-  const startInputRef = useRef<HTMLInputElement>(null);
-  const endInputRef = useRef<HTMLInputElement>(null);
-  const [focusedInput, setFocusedInput] = useState<FocusedInput>(
-    props.autoFocus ? "startDate" : null
-  );
-  const [isFocused, setIsFocused] = useState(props.autoFocus ?? false);
-  const [isOpen, setIsOpen] = useState(props.autoFocus ?? false);
-  const validationState = props.readOnly ? "notSet" : props.issues ? "invalid" : "valid";
+function dateToCalendarDate(date: Date): CalendarDate {
+  if (!date) return date;
```
gabro, line 35: I don't understand this line. In which case is `date` falsy and why is it ok to return a falsy value as `CalendarDate`?

Maybe we can use a more explicit check?

gabro (changes requested):

packages/bento-design-system/src/DateField/DateField.tsx
veej: I forgot to update it to use `Date`

packages/bento-design-system/src/DateField/DateField.tsx
veej: leftover

packages/bento-design-system/src/DateField/DateField.tsx
veej: Yes, in react-aria a range is defined only when both the start and end date are defined. When you start picking a range from the calendar, the start date in the input is not even updated until you pick also the end date. If you write the date manually in the input, the text you type is stored locally and the onChange is triggered only when you complete typing a valid date range.
So it's basically impossible to set only the start date. You have separate states for the selected range (null or complete range) and, eventually, the partial text typed manually in the inputs

gabro (approved):

PR #712, opened 2023-08-19
Enable turbosnap in Chromatic

gabro:


gabro:
Mm, this doesn't seem to work. Turbosnap is picking up changes to preview.tsx, probably because it's importing from stories/index.tsx which exports from bento.
I need to investigate

PR #732, opened 2023-08-25
Implement HomePage and MyTheme page

veej:
https://github.com/buildo/bento-design-system/assets/925635/9f514ed5-3b0e-495d-99f7-586ef0ef9091


packages/configuration-builder/src/ColorsSection/SectionCompleted.tsx
```diff
@@ -18,14 +16,14 @@ export function SectionCompleted(props: Props) {
           kind="solid"
           hierarchy="secondary"
           label={t("ColorsSection.returnToMyTheme")}
-          onPress={props.goToMyTheme}
+          onPress={() => navigate("/theme")}
```
gabro, line 19: `ButtonLink` instead of `Button` if we're navigating onPress?

packages/configuration-builder/src/ColorsSection/SectionCompleted.tsx
```diff
@@ -18,14 +16,14 @@ export function SectionCompleted(props: Props) {
...
         />
         <Button
           size="large"
           kind="solid"
           hierarchy="primary"
           label={t("ColorsSection.goToTypography")}
-          onPress={props.goToTypography}
+          onPress={() => navigate("/theme/typography")}
```
gabro, line 26: same

veej: fixed

gabro (approved):

gabro:
@veej there seems to be an issue with the lockfile. Can you run pnpm install locally and push the updated lockfile?

veej:
@gabro it seems the workaround to fix the chromatic flakiness with DateField is not working 😞

PR #749, opened 2023-08-29
Use Vite builder in Storybook

veej:


gabro (approved):

gabro (changes requested):
The Storybook build is failing due to a missing woff loader

veej:
@gabro moving the import of fonts in the `preview.ts` seems to fix the issue with the build

gabro (approved):

PR #751, opened 2023-08-29
Change renovate schedule

gabro:
Running renovate at any time is a bit annoying since it forces many rebases on our PRs.
This PR changes the behavior to:

- run Renovate only on Monday between 1am and 10am
- remove the hourly limit (but keeping the concurrent limit to 2)

This should allow us to update lots of dependencies all together once a week, without disrupting regular operations.


veej (approved):
❤️

PR #766, opened 2023-09-11
Fix backgrounds and colors

federico-ercoles:
This PR fixes some defects on background and font colors not being correctly aligned with Figma.

### DIFF
--- a/packages/bento-design-system/src/Field/Field.css.ts
+++ b/packages/bento-design-system/src/Field/Field.css.ts
@@ -12,11 +12,16 @@ export const inputRecipe = strictRecipe({
         color: vars.textColor.textSecondary,
       },
       selectors: {
+        "&:disabled, &[disabled]": {
+          background: vars.backgroundColor.backgroundPrimary,
+        },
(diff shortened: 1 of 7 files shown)


gabro:
@federico-ercoles changes look good, but the CI is complaining

packages/bento-design-system/src/Field/Field.css.ts
```diff
@@ -5,19 +5,25 @@ import { vars } from "../vars.css";
...
         "&:disabled::placeholder": {
           color: vars.textColor.textDisabled,
         },
-        "input&:read-only, textarea&:read-only, &.readOnly, &[readonly]": {
-          background: readOnlyBackground,
-        },
+        [`input&:read-only${notDisabled}, textarea&:read-only${notDisabled}, &.readOnly${notDisabled}, &[readonly]${notDisabled}`]:
+          {
+            background: readOnlyBackground,
+          },
```
federico-ercoles, line 26: @gabro  not sure if there's a better, less verbose way of doing this. The previous code was running correctly, but was generating build errors.

PR #769, opened 2023-09-14
Allow grouped table headers

federico-ercoles:
This PR adds support for grouped headers. The following options for simple headers are available:
* left-side stickiness: if a header group has `sticky: "left"`, the whole group will be sticky
* header stickiness
* label alignment
* hints


packages/bento-design-system/src/Table/Table.tsx
```diff
@@ -74,8 +81,11 @@ type SortingProps<C extends ReadonlyArray<ColumnType<string, {}, any>>> =
       onSort?: never;
     };
 
-type Props<C extends ReadonlyArray<ColumnType<string, {}, any>>> = {
-  columns: C;
+type Props<
+  C extends ReadonlyArray<SimpleColumnType<string, {}, any>>,
+  D extends ReadonlyArray<GroupedColumnType<string, {}, any>>
+> = {
```
federico-ercoles, line 87: @gabro I think this makes the PR breaking, but I'm not sure. What do you think?

PR #774, opened 2023-09-21
add option to virtualize Table rows

giogonzo:
This PR adds support for virtualizing the rows in a `Table`, using `@tanstack/react-virtual`. Virtualization is opt-in and for the time being it can't be enabled when grouping is also enabled


packages/bento-design-system/src/Table/Table.tsx
```diff
@@ -77,15 +78,18 @@ type SortingProps<C extends ReadonlyArray<ColumnType<string, {}, any>>> =
...
   noResultsDescription?: LocalizedString;
   noResultsFeedbackSize?: FeedbackProps["size"];
   initialSorting?: Array<SortingRule<C>>;
   stickyHeaders?: boolean;
   height?: { custom: string | number };
   onRowPress?: (row: Row<RowType<C>>) => void;
-} & SortingProps<C>;
+} & (
+  | { groupBy?: C[number]["accessor"]; virtualizeRows?: never }
+  | { groupBy?: never; virtualizeRows?: { estimateRowHeight: (index: number) => number } }
```
giogonzo, line 90: mutually exclusive for now as I didn't investigate how to support virtualizing tables with `groupBy` enabled

packages/bento-design-system/src/Table/Table.tsx
```diff
@@ -299,50 +320,67 @@ export function Table<C extends ReadonlyArray<ColumnType<string, {}, any>>>({
...
-      {rows.flatMap((row, index) => {
-        if (row.isGrouped) {
-          return [
-            <SectionHeader
-              key={row.groupByVal}
-              label={row.groupByVal}
-              numberOfStickyColumns={stickyLeftColumnsIds.length}
-            />,
-            ...row.leafRows.map((row, index) => {
+    <Box style={{ height: tableHeight(height) }} ref={tableContainerRef} overflow="auto">
```
giogonzo, line 323: Not 100% sure of the possible implications of adding an extra wrapper here

packages/bento-design-system/src/Table/Table.tsx
```diff
@@ -236,6 +241,29 @@ export function Table<C extends ReadonlyArray<ColumnType<string, {}, any>>>({
     }
   }, [data.length, stickyLeftColumnsIds]);
 
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+
+  const virtualizeRows =
+    typeof virtualizeRowsConfig === "boolean" ? virtualizeRowsConfig : virtualizeRowsConfig != null;
+  const estimateSize =
+    typeof virtualizeRowsConfig === "boolean"
+      ? () => 52
```
giogonzo, line 250: not sure how to pick a proper default value here. In practice, testing it, it doesn't seem to change much

gabro (approved):

PR #776, opened 2023-09-21
Fix withBentoConfig utility

federico-ercoles:
Locally the build is successful. Not sure if the fix is correct.


gabro (approved):
Let's try it!

PR #779, opened 2023-09-22
Allow virtualizing rows when groupBy is active

### LINKED ISSUE #778: Tech-debt: Support grouped rows in virtualized Table

giogonzo:
Closes #778

This PR adds the ability to render a Table that is both virtualized and uses the `groupBy` feature

### DIFF
--- a/packages/bento-design-system/stories/Components/Table.stories.tsx
+++ b/packages/bento-design-system/stories/Components/Table.stories.tsx
@@ -573,17 +573,34 @@ export const VirtualizedRows = {
   args: {
     stickyHeaders: true,
     height: { custom: 340 },
-    virtualizeRows: true,
+    virtualizeRows: { estimateRowHeight: () => 92 },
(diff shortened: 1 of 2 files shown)


packages/bento-design-system/stories/Components/Table.stories.tsx
```diff
@@ -573,17 +573,34 @@ export const VirtualizedRows = {
   args: {
     stickyHeaders: true,
     height: { custom: 340 },
-    virtualizeRows: true,
+    virtualizeRows: { estimateRowHeight: () => 92 },
```
giogonzo, line 576: not related to this PR, but I noticed that the table can glitch while scrolling if the estimated height is less than the true one. Since we are defaulting internally to the single line text cell height but with the example data we have multiple lines and thus higher rows, we have to provide the estimates from outside for the examples to work smoothly

gabro (approved):

PR #784, opened 2023-09-25
BEN-135: Tokens section (brand tokens)

veej:
https://github.com/buildo/bento-design-system/assets/925635/d4313ac9-b5a8-47b3-bf3b-708ec972dc40


packages/configuration-builder/src/TokensSection/BrandTokens.tsx
```diff
@@ -0,0 +1,104 @@
...
+                <Body size="medium" color="secondary">
+                  {t("Component.sliderField")}
+                </Body>
+                <SliderField
+                  type="single"
+                  label={unsafeLocalizedString("")}
+                  value={50}
+                  minValue={0}
+                  maxValue={100}
+                  onChange={() => {}}
```
gabro, line 95: minor, but I would add a valid `onChange` here. I've tried interacting with it and it felt "broken"

PR #804, opened 2023-10-19
Allow Table columns to have footers

federico-ercoles:


packages/bento-design-system/src/Table/Table.tsx
```diff
@@ -593,6 +618,60 @@ function ColumnHeader<D extends Record<string, unknown>>({
...
+        textAlign={column.align}
+        {...config.padding.footer}
+      >
+        <Box
+          paddingLeft={first ? config.boundaryPadding : undefined}
+          paddingRight={last ? config.boundaryPadding : undefined}
+        >
+          {column.Footer && (
+            <Columns space={8} alignY="center" align={column.align}>
+              {column.Footer ? (
```
veej, line 662: isn't this already checked by the `&&` at line 660? 
Also, do we want the wrapping Boxes if `column.Footer` is null?

veej (commented):
Just a small doubt

veej (approved):

PR #809, opened 2023-11-22
Fix config types for navigation.activeNavigation

gabro:
Our implementation of `PartialDeep` was not handling type unions correctly, so it would fail on `activeNavigation`.

I've fixed it by borrowing the type-fest's `PartialDeep` implementation (inlined with attributtion and license reference)


veej (approved):
👍

PR #814, opened 2023-11-27
Add text and select input

federico-ercoles:
This PR separates the input components from their field counterparts for `Text` and `Select`, similarly to what we already had for `Number` and `Checkbox`. Naming for these components was inferred from the existing examples.
I've had to refactor the `NumberInput` component, in order to align it with the new components, which I think would make the PR breaking.

I've done some basic testing in Storybook to confirm that there should be no visual or functional regression for the existing `Field` components, and also to confirm the functionality of the `Input` components.


veej (approved):
Looks 👍

PR #825, opened 2024-02-02
Add Table configs for header/footer size and default cells options

veej:


mperrando:
Hi @veej. 

Is it possible to add to the customisation of the header also the Icon of the hint? 

``` ts
const HintIcon = config.headerHintIcon ?? IconInfo
...
<HintIcon size={12} color="currentColor" />
```

maybe the whole element?

```ts
{config.headerHintElement ?? <IconInfo size={12} color="currentColor" />}
```

gabro (approved):
LGTM!

PR #828, opened 2024-02-21
Change style of sticky columns padding when using dividers

federico-ercoles:
When using a table with dividers, changed padding after sticky columns to be 2px wide and have for background the same color as the dividers.

Screenshot 2024-02-21 at 15 48 43


veej (approved):
Looks 👍

PR #859, opened 2024-06-26
Fix SelectField hydration mismatch error

gabro:
`SelectField` suffers from this react-select issue https://github.com/JedWatson/react-select/issues/2629 when rendering in SSR environment

This PR works around it by adding a stable id (using `useId` or the `id` passed externally) to pass to `id` and `instanceId`.


packages/bento-design-system/src/SelectField/BaseSelect.tsx
```diff
@@ -58,11 +58,16 @@ export function BaseSelect<A>(props: Props<A>) {
     clearable = true,
   } = props;
 
+  // NOTE(gabro): we want to make sure we have a stable ID across SSR rendering, to overcome this issue with react-select https://github.com/JedWatson/react-select/issues/2629
+  const generatedId = useId();
+  const id = fieldProps.id ?? generatedId;
```
chane81, line 63: 👍

PR #891, opened 2024-08-28
[Columns] Fix check for child as lazy node Column

### LINKED ISSUE #835: Bug: Column is not working with RSC
### Description of the bug

Trying to bootstrap a new Next.js application with App Router, and using `Columns` in a server component (e.g. a layout), it seems the `width` set on explicit `Column` components is ignored. E.g.

```
<Columns space={0}>
   <Column width="1/3">
       ....
   </Column>
   <Column width="2/3">
      ...
   </Column>
</Columns>
```
The code above will result in two `Column`s splitting the available space in half, instead of respecting the 1/3 - 2/3 proportion.

Inspecting the generated HTML code, it seems the `Column` component with the right flex-basis is wrapped in another `Column`, as if bento is not able to recognize the child is a `Column`:
image

It seems, for reasons, the `isColumn` function to detect if a child of `Columns` is a `Column` doesn't work as expected with RSC.

### Playroom reproduction

_No response_

### Version

0.22.9

Chrome

pedrodim:
Closes #835

Fix based on [this comment](https://github.com/buildo/bento-design-system/issues/835#issuecomment-2291839752). As the check now uses internals from a canary version of React any future updates to the payload format could break the check.

### DIFF
--- a/packages/bento-design-system/src/Layout/Columns.tsx
+++ b/packages/bento-design-system/src/Layout/Columns.tsx
@@ -1,4 +1,5 @@
 import { ReactChild, ReactElement } from "react";
+import { isLazy } from "react-is";
 import { flattenChildren } from "../util/flattenChildren";
 import { BoxProps, Box } from "../Box/Box";
 import { ResponsiveSpace } from "../internal";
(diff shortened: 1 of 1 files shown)


gabro (approved):
Code looks good! How did you test this?

gabro:
Good enough! Thanks

PR #908, opened 2024-09-18
Disclosure: config title spacing.

mperrando:


veej:
@mperrando Thank you so much for the PR 🙏 
I merged it into an internal branch, so that we can run automatic checks before merging on main

PR #922, opened 2024-10-16
Update to React 18.3.1 and fix two React warnings in Table

gabro:
Warning 1 is new to 18.3.x and is about spreading props that contain a `key`. We were doing it in Table because some react-table utilities return also a `key`.

Warning 2 is about using `useLayoutEffect` during server side rendering (see https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85). I've changed it to `useEffect` and it just works, so it was probably not necessary.


veej (approved):
💯

PR #925, opened 2024-10-18
Use flex positioning for SearchBar accessories to fix SSR glitches

gabro:
Not sure why I originally implemented this using absolute positioning, honestly...

The current implementation causes weird glitches in SSR applications, because the dimension of accessory is not known on the server.

Using normal flex positioning removes the problem and is also way less less convoluted


gabro:
Merging without a review since it seems trivial enough and I want to test it in an actual project

PR #988, opened 2025-03-13
Expose `DateInput` component

marcopiii:


packages/bento-design-system/src/DateField/DateField.tsx
```diff
@@ -109,17 +124,35 @@ function SingleDateField({ disabled, readOnly, ...props }: Extract<Props, { type
...
+  readOnly,
+  ...props
+}: Extract<InternalDateProps, { type: "range" }>) {
   const localTimeZone = getLocalTimeZone();
   const internalProps = {
     ...props,
     isDisabled: disabled,
     isReadOnly: readOnly,
-    validationState: props.issues ? "invalid" : "valid",
+    validationState: !props.isStandalone && props.issues ? "invalid" : "valid",
```
veej, line 155: Is this correct? Is there a specific reason to ignore the `validationState` when used as a standalone component?

packages/bento-design-system/src/DateField/DateField.tsx
```diff
@@ -109,17 +124,35 @@ function SingleDateField({ disabled, readOnly, ...props }: Extract<Props, { type
           shortcuts={shortcuts}
         />
       )}
+    </>
+  );
+
+  return props.isStandalone ? (
```
veej, line 130: can't we just expose the `datePicker`  as a `DateInput` component and then define the `DateField` as another component wrapping the DateInput?

veej (commented):
Thanks 🙏 
Just a couple of minor doubts inline

packages/bento-design-system/src/DateField/DateField.tsx
marcopiii: I did it because`validationState` is relevant only for the _field_ part, the standalone component doesn't use it

https://github.com/buildo/bento-design-system/blob/c5f26ca666fcd5c1ce0625ce9a27525d0f36b6c5/packages/bento-design-system/src/DateField/Input.tsx#L31-L43

Should we also add it to the standalone component? It would be coherent with `NumberInput` actually

packages/bento-design-system/src/DateField/DateField.tsx
marcopiii: That was my first approach, but the `useDatePicker` hook returns stuff that is necessary both for the _input_ and the _field_. If we separate them into two components, wouldn't we have to use `useDatePicker` both in `DateInput` and `DateField`?

packages/bento-design-system/src/DateField/DateField.tsx
veej: 🤔 I see the `internalProps` including it are passed only to `useDatePicker` and `useDatePickerState` which, according to the documentation, don't take a `validationState`. So I think it's irrelevant 👍

veej (approved):
👏

PR #992, opened 2025-03-17
#991: Expose `formatOptions` of `NumberInput`

### LINKED ISSUE #991: Expose `formatOptions` of `NumberInput`
The component `NumberInput` internally determines the number's format using the hook `useFormatOptions`.

https://github.com/buildo/bento-design-system/blob/e5751c166f868a548fc80fdadb7df9abd1101080/packages/bento-design-system/src/NumberField/formatOptions.ts#L4-L21

 The `kind` property defines it, and there is no way to alter it via the `NumberInput`'s API.

It would be very useful to be able to customize the format options.

marcopiii:
Closes #991


veej (approved):
👍

PR #999, opened 2025-04-04
merge config prop passed to BentoProvider with the one passed at creation

veej:


packages/bento-design-system/src/BentoProvider.tsx
```diff
@@ -88,11 +88,16 @@ export function createBentoProvider(
...
+        defaultMessages,
+      };
+    }, [defaultMessages]);
     return (
       <I18nProvider locale={locale}>
         <OverlayProvider style={{ height: "100%" }}>
-          <DefaultMessagesContext.Provider value={{ defaultMessages }}>
-            <BentoConfigProvider value={props.config ?? config}>
+          <DefaultMessagesContext.Provider value={defaultMessagesValue}>
+            <BentoConfigProvider value={deepmerge(config, props.config) as BentoConfig}>
```
pedrodim, line 100: I don't know how much overhead `deepmerge` adds when one object is `undefined`, but since `BentoConfigProvider` internally does a `deepmerge` too, maybe we can do the merge only if there's a `props.config`?

```suggestion
            <BentoConfigProvider value={props.config ? deepmerge(config, props.config)  as BentoConfig : config}>
```

veej: Good catch! Thank you 🙏

pedrodim (approved):

PR #1027, opened 2025-05-29
Add AsyncSelect

MalpenZibo:
Add support to load select options with a promise


packages/bento-design-system/src/SelectField/AsyncSelectField.tsx
pedrodim, line 1: This could be just:

```ts
import { LocalizedString, SelectOption } from "..";
import { SelectField, type SelectFieldProps } from "./SelectField";

type Props<A> = SelectFieldProps<A> & {
  loadOptions: (inputValue: string) => Promise<Array<SelectOption<A>>>;
  loadingMessage?: LocalizedString;
};

export type { Props as AsyncSelectFieldProps };

export function AsyncSelectField<A>(props: Props<A>) {
  return <SelectField {...props} />;
}
```

packages/bento-design-system/src/SelectField/components.tsx
```diff
@@ -296,6 +297,20 @@ export function MultiValue<A extends SelectOption<unknown>>(props: MultiValuePro
...
+    <Inset space={16}>
+      <Inline align="center" space={0}>
+        <InlineLoader message={props.children as LocalizedString} />
+      </Inline>
+    </Inset>
+  );
+}
+
+export function LoadingIndicator() {
+  return <></>;
```
pedrodim, line 311:
```suggestion
  return null;
```

pedrodim (changes requested):

packages/bento-design-system/src/SelectField/BaseSelect.tsx
```diff
@@ -66,81 +72,99 @@ export function BaseSelect<A>(props: Props<A>) {
...
         }
-        multiSelectMode={isMulti ? props.multiSelectMode : undefined}
-      />
+      }
+    },
+    onBlur: onBlur,
+    options: sortOptions(options),
+    placeholder: placeholder,
+    menuPortalTarget: menuPortalTarget.current,
+    components: selectComponents as any,
```
MalpenZibo, line 122: Here and on `styles` I had to use `as any`, and I haven't found a way to make the compiler happy.

pedrodim (approved):

PR #1091, opened 2026-07-29
feat: add configurable multi-select search persistence

giogonzo:
## What changed

- Add `dropdown.multiSelectSearchMode` with `clear-on-select` and `preserve-on-select` modes.
- Add an optional multi-select component prop that overrides the configured mode.
- Preserve searchable multi-select queries after selection while still clearing on Escape, menu close, genuine blur, and remount.
- Share the behavior between synchronous and asynchronous selects, reusing async results for a retained query.
- Add generic Storybook examples and focused interaction coverage.

## Why

Searchable multi-selects sometimes need to support consecutive selections from the same filtered result set. This adds that interaction as an opt-in mode while keeping the existing clear-on-select behavior as the default.
Discussed briefly with Livia in terms of behavior / UX and she seems to agree with this addition.

## Impact

Existing consumers are unchanged. The mode applies only to interactive searchable multi-selects; single-select and non-searchable components retain their current behavior. A component prop can opt in or override the provider-level default.

## Validation

- `pnpm prettier-check`
- `pnpm eslint-check`
- `pnpm check-circular-deps`
- `pnpm typecheck`
- `pnpm --filter @buildo/bento-design-system test -- --run` — 23 tests passed
- `pnpm build`
- `pnpm --filter @buildo/bento-design-system build-storybook`

## Human review

Tested via storybook (see added stories) and it is working as expected. Considering it is an optional config and the default stays the previous behavior, this should be safe to merge

### DIFF
--- a/packages/bento-design-system/test/SelectField.test.tsx
+++ b/packages/bento-design-system/test/SelectField.test.tsx
@@ -1,8 +1,7 @@
-import { render, screen, waitFor } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { useState } from "react";
 import {
-  AsyncSelectField,
   BentoProvider,
   PartialBentoConfig,
   SelectField,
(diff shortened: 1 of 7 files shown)


packages/bento-design-system/test/SelectField.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+    await user.type(input, "bl");
+    await user.click(screen.getByRole("button", { name: "Outside" }));
+    expect(input).toHaveValue("");
+
+    unmount();
+    renderWithConfig(<MultiSelect multiSelectSearchMode="preserve-on-select" />);
+    expect(screen.getByRole("combobox", { name: "Colors" })).toHaveValue("");
+  });
+
+  test("does not change single-select behavior when configured globally", async () => {
```
giogonzo: remove this and the following tests as they are superfluous / testing the config mechanism rather than this component behavior

veej (approved):

